PR #11, opened 2019-07-02
$Ifdef for inline keyword for support Delphi 7

fau:
D7 don't support "inline" keyword


andgineer:
https://stackoverflow.com/questions/42291399/does-delphi-7-support-the-inline-keyword

PR #12, opened 2019-07-08
Update RegExpr.pas

dbcto:
Support for Linux64 on Delphi 10.3.

### DIFF
--- a/src/RegExpr.pas
+++ b/src/RegExpr.pas
@@ -1,4 +1,4 @@
-unit RegExpr;
+﻿unit RegExpr;
 
 {
      TRegExpr class library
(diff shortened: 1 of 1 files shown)


andgineer:
could you please make tests green for your patch?
the tests run with Free Pascal but $DEF should keep this version good

"Can't find unit UnicodeData used by RegExpr"

Alexey-T:
this patch adds new WideUpperCase/WideLowerCase with LOOONG case operator. not good. better use FPC's widestrringmanager or Delphi's wide funcs.

Alexey-T:
And FPC already has the patch to support WideChar uppercase/ lowercase. so not needed.

dbcto:
> this patch adds new WideUpperCase/WideLowerCase with LOOONG case operator. not good. better use FPC's widestrringmanager or Delphi's wide funcs.

Yes, it's very LOOOONG ;) - but it's 5 times faster than System.SysUtils.UpperCase(Ch1)[1]. And the most important thing is the Linux compatibility for me for Delphi 10.3 which is not given with this implementation: uses
{$IFDEF WIN32}
    Windows; // CharUpper/Lower
{$ELSE}

andgineer:
Thank you!
But could you please fix the code so fpc(lazbuild) tests in travis CI will pass?

Alexey-T:
Please don’t apply this before my big patch

dbcto:
> Thank you!
> But could you please fix the code so fpc(lazbuild) tests in travis CI will pass?

Sorry - I have no fpc to check that and I didn't write the corresponding code in line 730. It seems to be from Alexey Fedorov.

andgineer:
@Alexey-T 
Could you please check if we can incorporate this Delphi change but still have your FPC optimisations?

Alexey-T:
You can close this PR, I will do this change using FPC and Delphi 2009+ Unicode funcs

Alexey-T:
@dbcto See #37 - standard uppercase/lowercase are needed to support all Unicode letters, your patch misses many letters

PR #23, opened 2019-11-13
Support non-capturing groups (?:text)

Alexey-T:


andgineer:
Thank you, great work!

Could you please add some test cases specific to the non-capturing work?

Alexey-T:
Added fix and test in Lazarus.

Alexey-T:
Screenshot from 2019-11-13 18-20-57

andgineer:
Thank you!
But why din't you just add cases into
https://github.com/andgineer/TRegExpr/blob/master/test/tcregexp.pas

Travis (.travis.yaml) use it to automatically test the project - see check results above and as a budge in README

Alexey-T:
Pls see again- added 2 testcases and fixed major bug in my diff.

src/RegExpr.pas
```diff
@@ -2576,7 +2576,9 @@ function TRegExpr.ParseAtom (var flagp : integer) : PRegExprChar;
            // check for non-capturing group: (?:text)
            if (regparse + 1)^ = ':' then begin
              if regnpar < NSUBEXP then
+             begin
```
andgineer: Why do we need begin/end for one liner?

test/tcregexp.pas
```diff
@@ -273,49 +274,56 @@   TRegExTest = record
...
 
 procedure TTestRegexpr.AreEqual(ArrorMessage: string; s1,s2: string);
 begin
-  {$IFDEF FPC}
-  AssertEquals(ArrorMessage, s1,s2);
-  {$ELSE}
-  CheckEquals(s1,s2, ArrorMessage)
-  {$ENDIF}
-end;
+  {$IFDEF FPC}
```
andgineer: not clear - what was changed?

test-non-capturing-groups/test-regex.lpi
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
andgineer: Do we still need this separate test project?
Now we have the test cases in main test suit

andgineer:
> Pls see again- added 2 testcases and fixed major bug in my diff.

Great work, thank you very much!

I just have a few minor questions about the changes

test-non-capturing-groups/test-regex.lpi
Alexey-T: Removed it.

test/tcregexp.pas
Alexey-T: It is some EOL changes made by Lazarus, or maybe trailing spaces deleted.

PR #24, opened 2019-11-14
change type of fInputString/fExpression from PChar to string, clearer code

Alexey-T:
… code

### DIFF
--- a/src/RegExpr.pas
+++ b/src/RegExpr.pas
@@ -295,8 +295,8 @@ type
     // to the thing following the set of BRANCHes.) The opcodes are:
     programm : PRegExprChar; // Unwarranted chumminess with compiler.
 
-    fExpression : PRegExprChar; // source of compiled r.e.
-    fInputString : PRegExprChar; // input string
+    fExpression : RegExprString; // source of compiled r.e.
+    fInputString : RegExprString; // input string
 
     fLastError : integer; // see Error, LastError
 
@@ -331,7 +331,6 @@ type
     // [re]compile it if something changed
     function IsProgrammOk : boolean; //###0.941
 
-    function GetExpression : RegExprString;
     procedure SetExpression (const s : RegExprString);
 
     function GetModifierStr : RegExprString;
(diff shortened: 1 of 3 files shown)


andgineer:
Sorry I do not understand what we need that for?
Inside there are still PChar processing - if looks for null-termination inside all routines.
So this change do not change the fact:
https://regex.sorokin.engineer/en/latest/faq.html#why-tregexpr-does-not-work-with-strings-with-0-inside

Alexey-T:
- it is good to simplify memory management - now not needed to do GetMem, FreeMem, ReallocMem, pascal does it auto. 
- next step will be supporting 0 chars

andgineer:
year I see.
you are right.

src/RegExpr.pas
```diff
@@ -1201,8 +1200,8 @@ constructor TRegExpr.Create;
  begin
   inherited;
   programm := nil;
-  fExpression := nil;
-  fInputString := nil;
+  fExpression := '';
```
andgineer: Are you sure this is safe to do?
I mean empty string ('') is not logically equivalent to not assigned (nil).
Of cause no sense to have empty expression but of cause you can empty input.
And even for expression - I just do not like this assumption 'nobody needs empty re'. What if..

Alexey-T: What is safe? setting empty str in Create? of course it's safe, and i even can delete this setting to empty str, because objects always have EMPTY str fields on creation

andgineer: My point we better distinguish between unassigned and assigned empty string.
Unassigned - error, empty string - no error.

Alexey-T: With pascal strings we cannot distinguish it, we can if we add a bool flag (var IsInited)... i am not sure we really need this flag to distinguish it, empty Expression / InputString can give the error- it's ok

Alexey-T: For ex, other Delphi components don't do this additional flag, e.g. Sql connection component just gives an error of connection string is empty

andgineer: wow :( ok thank you for clarification.
of cause if compiler cannot distinguish zero-length and unassingned we have no other choice

Alexey-T: Unassigned pascal string is a null pointer, when 1+ char is stored to string, it is not null

PR #28, opened 2019-11-15
Optimize speed of non-capturing groups storing

Alexey-T:


Alexey-T:
Seems it can solve all speed slowing, now its quite good

andgineer:
Great!
If only we had me performance test to measure the real impact..

PR #31, opened 2019-11-16
Support \v and \h, fix testcase for unicode mode

Alexey-T:


Alexey-T:
@andgineer Tasks with FPC 3.0 had passed ok. Tasks with FPC 2 will never pass- fpc 2.x missed unit "fpwidestring", so please delete fpc 2.x from CI.

andgineer:
You see TRegExpr now IS compatible with fpc 2.
All tests are green.
Why so minor change from you should break it?

Alexey-T:
@andgineer All checks passed now (after my fix to demo code)

andgineer:
Great solution thank you for your patience!

PR #34, opened 2019-11-22
Custom UpperCase/LowerCase which work ok on FPC and Delphi

Alexey-T:
this replaces PULL #12. it takes most code from it and adds some tweaks to that code, some optimizations to those uppercase/lowercase funcs.

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -5,12 +5,19 @@
                                      (+) - new feature
                                      (-) - bugfix
                                      (^) - improvement
+ v. 0.956 2019.11.22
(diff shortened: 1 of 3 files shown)


andgineer:
I am afraid the coversion table is much bigger
https://www.ibm.com/support/knowledgecenter/ssw_ibm_i_71/nls/rbagslowtoupmaptable.htm

And may be better idea just use system libraries for that?

Alexey-T:
Okay, i will use the IBM table from this page. could you apply the current patch so i can make next PULL to fix #35 ?

Alexey-T:
will make new PULL

PR #41, opened 2019-11-23
Allow NULL chars in InputString

Alexey-T:


Alexey-T:
@andgineer ?

andgineer:
Wow just great!
But don't you think the test is too simple?
I mean you add a lot of changes in different places but the test re is just simple string

Alexey-T:
А что ещё надо сделать в тесте? Нули я тестирую.

andgineer:
Your test do not visit most of this changed code.
Where all this "if" for different cases.

This is because you use just simple sting as re and code for char classes, groups and so on just does not work.

Alexey-T:
How do you suggest to make tests for my code? I don’t understand you

andgineer:
write more complex re

Alexey-T:
Ok, will do it

Alexey-T:
@andgineer Done

Alexey-T:
Note: \x{0} or \x00 don't work in regex, it cannot catch zero char and give ok result.

andgineer:
> Note: \x{0} or \x00 don't work in regex, it cannot catch zero char and give ok result.

Could you explain that - I do not understand, why we cannot use #0 inside regex now, when we do not treat it as string end.

Alexey-T:
- null is not handled in regex, it’s handled only in InputString (maybe later I try to do it in regex)
- why \x00 don’t work in character class- seems this writes null to op code, and later op code is read until null - Emit* functions look like it

andgineer:
thank you for clarification
may be we need ticket for that

PR #67, opened 2019-12-07
format regexpr.pas via Embarcadero Delphi 10.3 Formatter.exe, fix #65

### LINKED ISSUE #65: Format src
I can format source code a little, by mass replaces

- delete spaces before ( and [
- delete spaces before : and :=
- lowercase some keywords: Const Var
- title case EXIT

Ok? @andgineer

Alexey-T:


andgineer:
I do not like how IFDEF section now is formatted - the formatter removed all indentations and now it' hard to read

PR #71, opened 2019-12-08
Changes

Alexey-T:
- apply define UnicodeWordDetection also for Delphi2009+, set property to True by default
- format src: title case for Exit/Break/Continue
- fix place of define UnicodeWordDetection - define is now independent from Unicode
- del using define FPC_OS_UNICODE because fpc 3.2/3.3 don't have it on Linux x64

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -5,6 +5,13 @@
                                      (+) - new feature
                                      (-) - bugfix
                                      (^) - improvement
+ v. 0.962 2019.12.08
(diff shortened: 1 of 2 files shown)


andgineer:
May be we do need this safe guard
`{$IFNDEF UNICODE}
   {$UNDEF UnicodeWordDetection}
 {$ENDIF}`

in case somebody defines UnicodeWordDetection without  UNICODE?

Alexey-T:
Added it

Alexey-T:
@andgineer ?

PR #90, opened 2019-12-18
rework finding chars for char classes [], store to opcode pairs (Kind…

Alexey-T:
…,Data)

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -5,6 +5,11 @@
                                      (+) - new feature
                                      (-) - bugfix
                                      (^) - improvement
+ v. 0.966 2019.12.18
+ -=- (^) Reworked opcode for char-classes [], now opcode is stored as char pairs
+  (Kind,Data) without null-terminator, which allows to have null-char in char-classes,
+  and later have meta-chars like \W \D \S in char-classes. By Alexey Torgashin.
+
  v. 0.964 2019.12.12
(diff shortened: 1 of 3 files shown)


andgineer:
May we have to think about that more?
I am more concerned about additional comparison than the char-type itself.
We have to routinely compare types of each char. Why don't we have just type 'N chars next' and after that we just compare N-times with this chars and do not check type for each of them?

Alexey-T:
It can be optimized such a way (will give little), pls apply patch - I will then cleanup old code of Define UseCharSet, then will do optimization

andgineer:
I used set of chars 
https://www.freepascal.org/docs-html/ref/refsu16.html

for small chars number (less that 26) in hope this is good for performance.

of cause I cannot be sure without load tests.

did you have some reasons to remove that?

Alexey-T:
reason for removing:
- all this TinySet optimization was in the code which emitted null-terminated string, it was conflicting with new opcode (new opcode needs OpKind_Char+Len followed by a string). 
- this TinySet thing was used in NON unicode mode only
- i don't understand that TinySet code so cannot adapt it (it have ~6 variables, some functions)

Alexey-T:
failed CI test -- not my fail, see the report.

Alexey-T:
reason more:

- OP_ANYOFTINYSET cannot handle \w \W \s \S etc in charclasses. OP_ANYOF now can

Issue #123, opened 2020-01-06
CharCheckers for ranges


Alexey-T:
last Pullreq makes CharCheckers functions, and changed opcode for [] classes. why?
it's not faster yet. it will be faster later.
i will add opcodes (OpKind_) for more ranges:

a-z
A-Z
a-zA-Z
a-z0-9
A-Z0-9
a-zA-Z0-9
0-9 (will convert to OpKind_ for \d)

this will make faster parse on typical tests (test_benchmark demo)- it has many a-z ranges.
typical app often has such ranges.
CharCheckers are needed to make adding of these opcodes.

and we can add Checkers for more ranges:
а-я
А-Я
а-яА-Я

@andgineer

Alexey-T:
@andgineer 
new optimization, will be pull req.
compare of benchmark - 2 runs of old code (old1 old2), 2 runs of new code (new1 new2).
it shows that only one of tests is faster now (near the end), all others - the same.
```
user@user-Notebook:~/cuda/tregexpr/test_benchmark$ ./test1
                                                          old1       old2           new1          new2
=================================================================
RegExpr.pas:
                                        /Twain/ :        3374 ms     3297 ms        3292 ms       3313 ms 
                                    /(?i)Twain/ :        3430 ms     3395 ms        3399 ms       3412 ms 
                                   /[a-z]shing/ :        4279 ms     4274 ms        4269 ms       4262 ms 
                   /Huck[a-zA-Z]+|Saw[a-zA-Z]+/ :        3334 ms     3293 ms        3304 ms       3305 ms 
                                    /\b\w+nn\b/ :        4586 ms     4531 ms        4529 ms       4558 ms 
                             /[a-q][^u-z]{13}x/ :        6139 ms     6142 ms        6142 ms       6177 ms 
                  /Tom|Sawyer|Huckleberry|Finn/ :        3338 ms     3318 ms        3318 ms       3331 ms 
              /(?i)Tom|Sawyer|Huckleberry|Finn/ :        4116 ms     4067 ms        4100 ms       4086 ms 
          /.{0,2}(Tom|Sawyer|Huckleberry|Finn)/ :       17549 ms    35225 ms       17579 ms      18267 ms 
          /.{2,4}(Tom|Sawyer|Huckleberry|Finn)/ :       17655 ms    35217 ms       17705 ms      17880 ms 
[...]

PR #124, opened 2020-01-06
optimizations

Alexey-T:


Alexey-T:
@andgineer Last optim makes benchmark runs 3% faster

src/RegExpr.pas
```diff
@@ -3334,8 +3334,8 @@ function TRegExpr.ParseAtom(var flagp: integer): PRegExprChar;
           else
             Len := FindSkippedMetaLen(regparse + 1, fRegexEnd) + 1;
             // bad {n,m} - compile as EXACTLY
-        ender := (regparse + Len)^;
-        if (Len > 1) and ((ender = '*') or (ender = '+') or (ender = '?') or (ender = '{')) then
+        TempChar := (regparse + Len)^;
```
andgineer: For me old name was self-documented and new one is not so clear..

andgineer: > @andgineer Last optim makes benchmark run 3% faster

Great - this is scientific approach!

Alexey-T:
@andgineer Fixed var name to nicer: EnderChar.

Alexey-T:
@andgineer see #123 about last patch.

PR #126, opened 2020-01-07
move lower var declarations

Alexey-T:


src/RegExpr.pas
```diff
@@ -1575,9 +1574,9 @@ class function TRegExpr.InvertCaseFunction(const Ch: REChar): REChar;
 
 procedure TRegExpr.SetExpression(const AStr: RegExprString);
 begin
-  if (AStr <> fExpression) or not fExprIsCompiled then
+  if (AStr <> fExpression) or not regexpCompiled then
```
andgineer: in fact this is very strange looking - past form instead of passive

src/RegExpr.pas
```diff
@@ -2679,19 +2672,24 @@ function TRegExpr.ParsePiece(var flagp: integer): PRegExprChar;
...
       PREBracesArg(AlignToInt(Result + REOpSz + RENextOffSz))^ := ABracesMin;
-      PREBracesArg(AlignToInt(Result + REOpSz + RENextOffSz + REBracesArgSz))^
-        := ABracesMax;
+      PREBracesArg(AlignToInt(Result + REOpSz + RENextOffSz + REBracesArgSz))^ := ABracesMax;
     end;
   end;
 
+var
+  op: REChar;
+  NonGreedyOp, NonGreedyCh: boolean; // ###0.940
```
andgineer: Sorry I do not get it - somewhere it starts from capital letter, somewhere else this is cameCase.

src/RegExpr.pas
```diff
@@ -3492,15 +3490,15 @@ function TRegExpr.regrepeat(p: PRegExprChar; AMax: integer): integer;
         ArrayIndex := FSubExprIndexes[Ord(opnd^)];
         if ArrayIndex < 0 then
           Exit;
-        sestart := startp[ArrayIndex];
-        if sestart = nil then
+        GrpStart := startp[ArrayIndex];
```
andgineer: Well it's ok to rename subexpressions to groups.
But can we do that in all places? Now you have FSubExprIndexes and GrpStart

andgineer:
I can merge it if you like it.
But don't you think this is inconsistent in some places?

src/RegExpr.pas
Alexey-T: renamed to regexpIsCompiled, ok.

src/RegExpr.pas
Alexey-T: NonGreedyOp / NonGreedyCh are your old names, i moved them lower to make them not visible by included funcs.

andgineer:
Can we have some test for this (that 'foobar{3,blahblah' raises exception)?

src/RegExpr.pas
Alexey-T: renamed to GrpIndexes/ GrpCount, okay.

Alexey-T:
> Can we have some test for this (that 'foobar{3,blahblah' raises exception)?

@andgineer Test cannot show ok for property Error raised, so i tested in test_dlg program: now bad braces give error.

andgineer:
Well all tests as different soubroutines so we can have special subroutine for this test

Alexey-T:
Don't know how to write this- if like this, we don't get error msg at all
```
procedure TTestRegexpr.TestBadBraces;
begin
  try
    CompileRE('d{');
    IsTrue('Not raised Error for braces', False);
  except
  end;

  try
    CompileRE('d{2,');
    IsTrue('Not raised Error for braces', False);
  except
  end;

  try
    CompileRE('d{}');
    IsTrue('Not raised Error for braces', False);
  except
  end;
end;
```

PR #134, opened 2020-01-09
optimizations

Alexey-T:


src/RegExpr.pas
```diff
@@ -4247,20 +4242,26 @@ function TRegExpr.MatchAtOnePos(APos: PRegExprChar): boolean;
...
-    startp[i] := nil;
-    endp[i] := nil;
-  end;
+    GrpIndexes[i] := -1;
+  GrpIndexes[0] := 0;
+  GrpCount := 0;
 end;
 
-function TRegExpr.ExecPrim(AOffset: integer; ATryOnce: boolean): boolean;
+function TRegExpr.ExecPrim(AOffset: integer; ATryOnce, ASlowChecks: boolean): boolean;
```
andgineer: Looks like you just switch off optimization completely
You use ASlowChecks = False and do not use regmustString in this case.
BTW this is contr-intuitive - we slow check if ASlowCheck is False.

Alexey-T:
Yes, so I added Exec with param ASlowChecks- user can set True. Speed depends on user data size - if user uses 2M string the better have slow checks off. If he uses short buffers, better have slow checks on. Make autodetection? Not good idea. Exec overload - good

andgineer:
Why don't you like autodetection?
We can have some const - min input text size when we use this optimization.
May be that would be more understandable for a library user that ASlowCheck?

Alexey-T:
Ok, i will add property, with value e.g. 1000, if InputString is bigger than slowchecks off.

slowcheck will be off:

- on ExecNext call (because first Exec call already found regmustString)
- on Exec with Offset>1 (because it may be search from buffer middle)

slowcheck will be on if Offset=1 and Length(InputString)< new_prop. ok?

andgineer:
Great.
Of cause if Offset=1 and Length(InputString) > new_prop

And may bigger value - 1M? Could you compare speed for different input size where it actually matter?

Alexey-T:
Patch added.
i cannot compare speed, test_benchmark has buffer size 19M (Twain file), it will be slower of coz.

Alexey-T:
?

src/RegExpr.pas
```diff
@@ -4204,34 +4199,39 @@ function TRegExpr.MatchPrim(prog: PRegExprChar): boolean;
...
 
 {$IFDEF OverMeth}
 function TRegExpr.Exec: boolean;
+var
+  Len: integer;
+  SlowChecks: boolean;
 begin
-  Result := ExecPrim(1, False);
+  Len := Length(fInputString);
+  SlowChecks := (Len >= fSlowChecksSizeMin) and (Len < fSlowChecksSizeMax);
```
andgineer: Why do we need min?
We do not want to use optimisation on too small (Len < fSlowChecksSizeMax) -ok.
But why DO we use optimisation on very small (for Len < fSlowChecksSizeMin SlowCheck will be OFF so we will use optimisation)?

src/RegExpr.pas
```diff
@@ -1523,6 +1529,9 @@ constructor TRegExpr.Create;
   FUseUnicodeWordDetection := True;
   {$ENDIF}
 
+  fSlowChecksSizeMin := 20;
+  fSlowChecksSizeMax := 2000;
```
andgineer: May be we need some comments here?

Alexey-T: Added comment

Alexey-T:
Deleted Min property

Issue #147, opened 2020-05-07
Valid regex throws metachar after "-" in [] range


SlMaker:
valid regex (see regex101):
```
[.-]
```
error:
> metachar after "-" in [] range

Worked fine in older version.

Alexey-T:
you are doing smth wrong. CudaText works with this regex
![Screenshot from 2020-05-11 01-55-49](https://user-images.githubusercontent.com/4154978/81512662-e9e17d80-932a-11ea-8874-dd47af837782.png)

SlMaker:
```pascal
program TRegExprTest;

{$ifdef FPC}
{$mode Delphi}
{$else}
{$APPTYPE CONSOLE}
{$endif}

uses
  SysUtils,
  RegExpr in 'TRegExpr-master\TRegExpr-master\src\RegExpr.pas';

function TestRegex(input, expr: String): boolean;
var
  rm: TRegExpr;
begin
    rm := TRegExpr.Create;
    rm.ModifierI := False;
    rm.Expression := expr;
    Result := rm.Exec(input);
end;

begin
  try
    writeln (BoolToStr(TestRegex('This.Is.A.Test.1234', '[.-](Test)[.-]'), True));
    writeln (BoolToStr(TestRegex('This.Is.A.Test.1234', '[.\-](Test)[.\-]'), True));
    writeln (BoolToStr(TestRegex('This.Is.A.Test.1234', '[\.\-](Test)[\.\-]'), True));
    readln;
  except
    on E: Exception do
      Writeln(E.ClassName, ': ', E.Message);
  end;
end.
```
prints
```
False
True
True
```
which shows that it doesn't work.

Alexey-T:
I am not sure how such regexes should work, sorry

hrxcodes:
All the try cases from @SlMaker used to evaluate to `True` before

PCRE / ECMAScript / Python and GoLang implementations has all of the regexes in the try block working as I would expect and TRegExpr did evalute these to `True` before I updated the lib and there is no such regression mentioned.

## Examples
* https://regex101.com/r/gtS7Ue/1
* https://regex101.com/r/gtS7Ue/2
* https://regex101.com/r/gtS7Ue/3


## External info

Below info is copied from [this link](https://www.regular-expressions.info/charclass.html)
[...]

PR #150, opened 2020-05-11
Add tests for specific valid regex set

SlMaker:
#147


SlMaker:
@andgineer that one should be fine now

SlMaker:
@Alexey-T for me both regex with `[.-]` are failing

andgineer:
`([.-])` you have to escape `.` if you want a dot itself and not meta-char "any symbol"
`([\.-])`

Issue #146, opened 2020-05-07
Invalid regex are supported and working


SlMaker:
Working regexes which should fail:
```
[._-\s]((19|20)\d{2})
[._-\s](US|UK|AU|CA|NZ)[\s._-]
```
regex101: 
> You can not create a range with a shorthand escape sequences

Shouldn't allow the **-** at that position in the regex.

SlMaker:
@Alexey-T

Alexey-T:
Maybe i can add the error on such case. what @andgineer thinks?

masterandrey:
“-]” is ok this is shorthand to add “-“ in the range without backslash.
“-\s]” has no sense that should be an error. You can express that as “\s-]” or “\-\s]”

Alexey-T:
@masterandrey agree.

Alexey-T:
@SlMaker you are doing smth wrong
here with test.pas with
```
procedure TTestRegexpr.TestBads;
begin
  //TestBadRegex('No Error for bad braces', 'd{');
  //TestBadRegex('No Error for bad braces', 'd{22');
  //TestBadRegex('No Error for bad braces', 'd{}');
  TestBadRegex('Invalid char class after dash in []', '[._-\s]((19|20)\d{2})');
end;
```
I got FPC error
![Screenshot from 2020-05-11 01-49-10](https://user-images.githubusercontent.com/4154978/81512564-1943ba80-932a-11ea-91ec-98af2d271e5e.png)

SlMaker:
Was accepted and working in older versions but doesn't work any more but is correct

PR #151, opened 2020-05-11
Invalid regex

SlMaker:
#146 which was working in all prior versions but is fixed now


andgineer:
@SlMaker you can uncomment the test now?

SlMaker:
@andgineer 
the regex should never work as it is not valid at all (but older TRegExpr versions allowed it). Don't know how to test invalid regexes with your testsuite...

andgineer:
invalid re you test exactly as in you PR - TTestRegexpr.TestBadRegex
and in fact last one should raise an exception now..

SlMaker:
Why are the others uncommented then? Should I enable them as well?

Alexey-T:
they are commented. to not raise errors in test project. on testing I temporary enable em.

PR #167, opened 2020-06-11
define FastUnicodeData to speedup unicode checks

Alexey-T:


Alexey-T:
Comparing
before
bench-old-uni

after
bench-new-uni

NOTE- ASCII test- is faster, Uni will be more faster!

andgineer:
project1, unit1?
holiday crossword puzzle? ;)

Alexey-T:
Fixed, renamed project make_unidata files

andgineer:
thank you!

PR #183, opened 2020-08-18
Unicode categories \p \P

Alexey-T:
Tested yet only in Dialog demo, no separate tests.


andgineer:
May be we also need "infinity" value for MaxRegexBackTracking ?

Alexey-T:
I removed code which uses  MaxRegexBacktracking, so it's not used const.

andgineer:
@Alexey-T 
Agree. We could have just explicit exception saying that the engine have stopped due to this reason and you can change this cont but I am not sure it worths effort..

PR #222, opened 2020-09-04
opcode.txt

Alexey-T:


andgineer:
May be we better add this into docs as separate section for potential contributors?

Alexey-T:
Hmmm, too much troubles for me..

PR #233, opened 2020-11-29
Fix build with Delphi 10.4.1 when inlining is enabled.

zedalaye:
Avoid this bug error : http://docwiki.embarcadero.com/RADStudio/Sydney/en/E2441_Inline_function_declared_in_interface_section_must_not_use_local_symbol_%27%25s%27_(Delphi)

I had to move all OP_* variables up to the interface section

And:
http://docwiki.embarcadero.com/RADStudio/Sydney/en/E2449_Inlined_nested_routine_%27%25s%27_cannot_access_outer_scope_variable_%27%25s%27_(Delphi)

I had to explicitely pass the instance of GrpIndexes to FindSubstGroupIndex() and while I was at it I created types for all array [0 .. RegexMaxGroups - 1] of Something

### DIFF
--- a/src/regexpr.pas
+++ b/src/regexpr.pas
@@ -241,19 +241,24 @@ type
   end;
   TRegExprCharCheckerInfos = array of TRegExprCharCheckerInfo;
 
+  TRegExprGrpFlags = array [0 .. RegexMaxGroups - 1] of boolean;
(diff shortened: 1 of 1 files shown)


Alexey-T:
maybe better do this? unmark funcs which are 'inline' to make them normal. to make Delphi no errors. do it with IFDEF for Delphi.

Alexey-T:
@andgineer please close, solved.

PR #235, opened 2020-12-27
Remove wrong inlines directives which break Delphi compatibility

SlMaker:
FPC is not strict enough but that could change any time.
FPC devs also say that the inline directive should not be used as the compiler knows better when to inline and when not


Alexey-T:
@andgineer Pls apply this if user had it tested (I didnt test, no Delphi here)

PR #242, opened 2021-02-19
* Fix compilation for D2009> when no FastUniCodeData is defined

mvancanneyt:
Patch for issue 241 as requested.


Alexey-T:
@mvancanneyt Pls revert this 
`{ $DEFINE FastUnicodeData} `
because we need FPC version to be run with FastUnicodeData;
or you may add this

{$IFDEF FPC}
{$DEFINE FastUnicodeData}
{$ENDIF}

mvancanneyt:
Pushed as requested

Alexey-T:
@andgineer Pls apply?

Alexey-T:
@mvancanneyt 
`{$IFDEF FPC} // Not supported in FreePascal`
hmm

PR #243, opened 2021-02-19
Studio fix

mvancanneyt:
If you want, here are the fixes for compiling REStudio under Lazarus and Delphi.
All seems to work.

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -6,6 +6,10 @@
                                      (-) - bugfix
                                      (^) - improvement
 
+ v. 1.149 2021.01.08
+ -=- (-) Allow underscore char to be escaped '\_'. Other word-chars are
+ not allowed. By Alexey Torgashin.
+
  v. 1.148 2020.09.29
  -=- (-) Fixed broken ^ and $ if UseLinePairedBreak=False.
   By Alexey Torgashin.
--- a/History.txt
+++ b/History.txt
@@ -6,6 +6,9 @@
                                      (-) - bugfix
                                      (^) - improvement
 
+ v. 1.150 2021.02.08
+ -=- (-) Setters of Modifier* properties now invalidate the opcode.
+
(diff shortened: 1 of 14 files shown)


Alexey-T:
Only @andgineer can check it(I don't use Studio)

Alexey-T:
@mvancanneyt just a guess. if you add

{$IFDEF FPC}
{$mode delphi}
{$ENDIF}

it will help with `^` issue? then we dont need NODEREF define?

mvancanneyt:
No. 
The in-memory structure of RTTI in FPC is different. This has nothing to do with the mode.
Worse, it will change more in the future. So the define is really needed.

mvancanneyt:
By the way, I just contributed an article to Blaise Pascal magazine: regex for beginners. 
Scheduled to appear in the next issue.

TRegexpr and REStudio for Delphi/lazarus are demonstrated, so perhaps it is a good idea to clean up REStudio a little, so at least it works out of the box. :-)

andgineer:
@Alexey-T 
How do you think about changes in regexpr.pas ?
Is it safe to merge?

Alexey-T:
@andgineer 
This fix conflicts with the current code:
```
uses
  Classes, // TStrings in Split method
  SysUtils, // Exception
  {$IFDEF D2009}
    {$IFDEF D_XE}System.{$ENDIF}Character,
  {$ENDIF}
  Math;
```
so apply, but without regexpr.pas.

andgineer:
Sorry I give up at least for a moment.
I revert regexpr.pas from initial for this PR commit and even squashed commits to have just one change but still github show the same changes as in master as changes in this commit.
I am not sure what will happen after merge so I better won't merge it.
May be we should revert regexpr.pas changes in @mvancanneyt fork - I am not sure.

Issue #244, opened 2021-03-11
compiling errors on Delphi 2010


mirkasp:
Hi,
There are errors when compiling on Delphi 2010:

[DCC Fatal Error] regexpr.pas(116): F1026 File not found: 'System.Character.dcu'
and
[DCC Error] regexpr.pas(859): E2441 Inline function declared in interface section must not use local symbol 'PtrPair'

Best regards
MirKasp

Alexey-T:
Is unit System.Character in D2010? 
http://docwiki.embarcadero.com/RADStudio/Sydney/en/Character_Manipulation_Routines

it is in D 10.4.
in what D did this unit appear?

mirkasp:
Yes,  but without prefix "System." - only "Character.pas" 
Only two files have a prefix: "Generics.Collections.pas" and "Generics.Defaults.pas". 
All other source files have single-member names, for example: StrUtils.pas, Classes.pas, AnsiStrings.pas, etc.

Alexey-T:
Can you make the easy fix, and post your Pull req? I dont have Delph now.

andgineer:
@mirkasp Is the problem solved?
could you please close the ticket?

mirkasp:
Yes, it's work !

PR #245, opened 2021-03-11
regexpr.pas fix for Delphi 2010 

mirkasp:
Small fix for issue #244 
Not tested under other Delphi version.


Alexey-T:
@andgineer we can apply but without 2nd part (refactoring for IFDEF Unicode).
No reason to refactor here.

andgineer:
@mirkasp 
Could you please change PR according to Alexey's comment?

Alexey-T:
@andgineer Pls close w/o applying, the Delphi part of this patch was fixed already.

Issue #299, opened 2023-08-14
Allow to set fInputStart and fInputEnd


User4martin:
An application may have a long string, and needs to apply a pattern to a substring, such that `^` and `$` will match begin/end of the substring.

Currently the application must make a copy of the substring  `RegEx.InputString := copy(text, a, b);` 
It would be nice to save the time of copying strings around. 

Alternatively the TRegEx could allow to set InputString from a PChar /PWideChar. And specify the length (instead of looking for a terminating #0)

Alexey-T:
we have private methods
```
procedure TRegExpr.SetInputString(const AInputString: RegExprString);
begin
  ClearMatches;

  fInputString := AInputString;
  UniqueString(fInputString);

  fInputStart := PRegExprChar(fInputString);
  fInputEnd := fInputStart + Length(fInputString);
end;

procedure TRegExpr.SetInputRange(AStart, AEnd: PRegExprChar);
begin
  fInputString := '';
  fInputStart := AStart;
  fInputEnd := AEnd;
end;          
```

so i suggest to add SetInput public method:
SetInout(const AInoutString: RegExprString; AStartPos, ALen: integer);
can you make the patch and test it?

pls use our current coding style.

User4martin:
I have a look...

Just started to see, how much `FInputString` is used...

Would I be right to say that
```
function TRegExpr.ExecPrim(AOffset: integer;
  ATryOnce, ASlowChecks, ABackward: boolean): boolean;
begin
...
  Ptr := fInputStart + AOffset - 1;

  // If there is a "must appear" string, look for it.
  if ASlowChecks then
    if regMustString <> '' then
      if Pos(regMustString, fInputString) = 0 then Exit;
```

the `Pos(regMustString, fInputString)` is wrong (inefficent)?

Unless `regLookbehind = true` (and even then I don't know if 'regMustString' can be in a look-behind), then the `regMustString` must appear after AOffset?

Alexey-T:
>the regMustString must appear after AOffset?

yes, it looks like a bug. thanks. can you pls, prepare the fix? i can fix but no time to test all stuff.

User4martin:
I don't know so, if the `regMustString ` can have been found in a look behind.
```
          if PREOp(scan)^ = OP_EXACTLY then
```
I.e. can `OP_EXACTLY` be found if the text is in a look behind...

So to be carefully, if there are lookbehind, the current behaviour needs to be kept.

----

[...]

PR #300, opened 2023-08-15
Implement `SetInputSubString`. Issue #299

User4martin:
Please review. / See issue #299

1) I wasn't sure if `property InputString` should be modified, so it only returns the sub-string?

2) Since `fInputString` may contain extra text, I replaced all references to it.

3) I added `strLpos` for the lookup in p(wide)char, with a given max len. To further optimize I can add a `strLScan` (copied from fpc sources), and amend it to have a length param. Currently it simply checks the result, if it has gone to far. 

1 & 3 => depend on feedback.

### DIFF
--- a/src/regexpr.pas
+++ b/src/regexpr.pas
@@ -6975,7 +6975,7 @@ begin
     exit;
 
   fInputString := AInputString;
-  UniqueString(fInputString);
+  //UniqueString(fInputString);
 
   fInputStart := PRegExprChar(fInputString) + AInputStartPos - 1;
   fInputEnd := fInputStart + AInputLen;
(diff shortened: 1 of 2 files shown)


User4martin:
I did change the name (from the proposal in the issue), to make the purpose more clear. Also to avoid conflict with any future functionality.

Alexey-T:
> p:=strscan(str1,str2^);

this is slower than needed. can you add scanning of needed len only?

>   Dec(AOffset);

you added this, pls don't do it. we can get hard to catch bugs. if AOffset is used below and we won't fix the code.

> UniqueString(fInputString);

not needed, ok?


Also, not required, but the main author is me, so let's do PR to my repo instead?
when all is settled, i will do final PR to 'off repo'.

Alexey-T:
Seems OK now. @andgineer

Alexey-T:
PR is still needed?

Issue #297, opened 2023-08-14
Feature \G please


User4martin:
Could \G be implemented please?

https://www.tutorialspoint.com/regular-expression-g-metacharacter-in-java

Also  `ATextStartOffset` in   `FRegEx.Exec(ATextStartOffset)`

User4martin:
@Alexey-T  is this something you could help, please?

Alexey-T:
Maybe ... but I did not understand the text about \G. 
any info on your link.
and on this link- https://stackoverflow.com/questions/14897949/what-is-the-use-of-g-anchor-in-regex

Alexey-T:
first answer here
https://stackoverflow.com/questions/14365969/how-the-anchor-z-and-g-works-in-ruby/14366062#14366062
is good, but I still don't understand this idea.

Alexey-T:
OK I got the idea. after some time.
is is possible in engine - will think...

Alexey-T:
Martin, it is the easy change. i hope you can make the PR. its easy to remember the end of the match after successfull Exec* call, in the new variable, and it's easy to add the `\G` which handler must only compare the current-pos with that variable

I forgot the code, but here is example
```
    '^':
     begin
      if not fCompModifiers.M
        {$IFDEF UseLineSep} or (fLineSeparators = '') {$ENDIF} then
        ret := EmitNode(OP_BOL)
      else
        ret := EmitNode(OP_BOLML);
     end;

    '$':
     begin
      if not fCompModifiers.M
        {$IFDEF UseLineSep} or (fLineSeparators = '') {$ENDIF} then
        ret := EmitNode(OP_EOL)
      else
        ret := EmitNode(OP_EOLML);
     end;
```

so, add OP_PREVIOUS_END const. emit this code like this
```
          'A':
            ret := EmitNode(OP_BOL);
          'z':
            ret := EmitNode(OP_EOL);
```

Alexey-T:
@User4martin Let's close?

PR #303, opened 2023-08-15
Implement \G #297

User4martin:
See #297

This is based on #300 as some changes affect nearby lines and lead to conflicts if individually applied.

Once #300 is merged, only one commit should remain to be merged.

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -6,6 +6,11 @@
                                      (-) - bugfix
                                      (^) - improvement
 
+ v. 1.161 2023.08.17
+ -=- (-) Fix atomic groups. Allow backtracking the entire group.
+  Make '*','+',... operators aware of the atomic behaviour. Allow nesting.
+  By Martin Friebe.
+
  v. 1.160 2023.08.15
  -=- (+) Add \G for ExecNext to match the position where the previous match ended.
   Or in Exec(AOffset) match the AOffset position
--- a/History.txt
+++ b/History.txt
@@ -6,6 +6,14 @@
                                      (-) - bugfix
                                      (^) - improvement
 
+ v. 1.160 2023.08.15
+ -=- (+) Add \G for ExecNext to match the position where the previous match ended.
+  Or in Exec(AOffset) match the AOffset position
+  By Martin Friebe and Alexey Torgashin.
+ -=- (+) Add SetInputSubString(Text, Start, Len)
+  to only match within the selected substring.
+  By Martin Friebe and Alexey Torgashin.
+
  v. 1.158 2022.05.30
  -=- (-) Fix Substitute() result when 'subroutine' (?1) is used.
  -=- (-) Fix work of recursion (?R) with back-references.
--- a/docs/regular_expressions.rst
+++ b/docs/regular_expressions.rst
@@ -193,7 +193,7 @@ Meta-char     Matches
 ``\A``        zero-length match at the very beginning
 ``\z``        zero-length match at the very end
 ``\Z``        like ``\z`` but also matches before the final line-break
-``\G``        zero-length match at the end pos of the previous match.
+``\G``        zero-length match at the end pos of the previous match
 ============= ================================================
 
 Examples:
--- a/docs/regular_expressions.rst
+++ b/docs/regular_expressions.rst
@@ -193,6 +193,7 @@ Meta-char     Matches
 ``\A``        zero-length match at the very beginning
 ``\z``        zero-length match at the very end
 ``\Z``        like ``\z`` but also matches before the final line-break
+``\G``        zero-length match at the end pos of the previous match. For ExecNext and Exec(Offset)
 ============= ================================================
 
 Examples:
--- a/docs/regular_expressions.rst
+++ b/docs/regular_expressions.rst
@@ -193,7 +193,7 @@ Meta-char     Matches
 ``\A``        zero-length match at the very beginning
 ``\z``        zero-length match at the very end
 ``\Z``        like ``\z`` but also matches before the final line-break
-``\G``        zero-length match at the end pos of the previous match. For ExecNext and Exec(Offset)
+``\G``        zero-length match at the end pos of the previous match.
 ============= ================================================
 
 Examples:
--- a/src/regexpr.pas
+++ b/src/regexpr.pas
@@ -283,7 +283,7 @@ type
     GrpIndexes: array [0 .. RegexMaxGroups - 1] of integer; // map global group index to _capturing_ group index
     GrpNames: array [0 .. RegexMaxGroups - 1] of RegExprString; // names of groups, if non-empty
     GrpAtomic: array [0 .. RegexMaxGroups - 1] of boolean; // group[i] is atomic (filled in Compile)
-    IsBacktrackingGroupAsAtom: Boolean;  // Backtracing an atomic group that has matched.
+    IsBacktrackingGroupAsAtom: Boolean;  // Backtracking an entire atomic group that had matched.
     // Once the group matched it should not try any alternative matches within the group
     // If the pattern after the group fails, then the group fails (regardless of any alternative match in the group)
 
--- a/src/regexpr.pas
+++ b/src/regexpr.pas
@@ -1871,7 +1871,7 @@ begin
   if (AStr <> fExpression) or not IsCompiled then
   begin
     fExpression := AStr;
-    UniqueString(fExpression);
+    //UniqueString(fExpression);
     fRegexStart := PRegExprChar(fExpression);
     fRegexEnd := fRegexStart + Length(fExpression);
     InvalidateProgramm;
@@ -5770,7 +5770,7 @@ begin
   ClearMatches;
 
   fInputString := AInputString;
-  UniqueString(fInputString);
+  //UniqueString(fInputString);
 
   fInputStart := PRegExprChar(fInputString);
   fInputEnd := fInputStart + Length(fInputString);
--- a/src/regexpr.pas
+++ b/src/regexpr.pas
@@ -6975,7 +6975,7 @@ begin
     exit;
 
   fInputString := AInputString;
-  UniqueString(fInputString);
+  //UniqueString(fInputString);
 
   fInputStart := PRegExprChar(fInputString) + AInputStartPos - 1;
   fInputEnd := fInputStart + AInputLen;
--- a/src/regexpr.pas
+++ b/src/regexpr.pas
@@ -996,27 +996,25 @@ end;
 
 
 function StrLPos(str1,str2 : PRegExprChar; len1, len2: SizeInt) : PRegExprChar;
-  var
-    p : PRegExprChar;
-  begin
-    StrLPos := nil;
-    if (str1 = nil) or (str2 = nil) then
-      exit;
-    len1 := len1 - len2 + 1;
-    p := StrLScan(str1,str2^, len1);
-    if p = nil then
-       exit;
-    while p <> nil do
-      begin
-        if strlcomp(p, str2, len2)=0 then
-          begin
-             StrLPos := p;
-             exit;
-          end;
-        inc(p);
-        p := StrLScan(p, str2^, len1 - (p-str1));
-      end;
-  end;
+var
+  p : PRegExprChar;
+begin
+  StrLPos := nil;
+  if (str1 = nil) or (str2 = nil) then
+    exit;
+  len1 := len1 - len2 + 1;
+  p := StrLScan(str1,str2^, len1);
+  while p <> nil do
+    begin
+      if strlcomp(p, str2, len2)=0 then
+        begin
+           StrLPos := p;
+           exit;
+        end;
+      inc(p);
+      p := StrLScan(p, str2^, len1 - (p-str1));
+    end;
+end;
 
 {$IFDEF FastUnicodeData}
 function _UpperCase(Ch: REChar): REChar; {$IFDEF InlineFuncs}inline;{$ENDIF}
--- a/src/regexpr.pas
+++ b/src/regexpr.pas
@@ -427,6 +427,7 @@ type
     function DumpCategoryChars(ch, ch2: REChar; Positive: boolean): RegExprString;
 
     procedure ClearMatches; {$IFDEF InlineFuncs}inline;{$ENDIF}
+    procedure ClearInternalExecData; {$IFDEF InlineFuncs}inline;{$ENDIF}
     procedure ClearInternalIndexes; {$IFDEF InlineFuncs}inline;{$ENDIF}
     function FindInCharClass(ABuffer: PRegExprChar; AChar: REChar; AIgnoreCase: boolean): boolean;
     procedure GetCharSetFromCharClass(ABuffer: PRegExprChar; AIgnoreCase: boolean; var ARes: TRegExprCharset);
@@ -5590,6 +5591,7 @@ begin
     if (APos - fHelperLen) >= fInputStart then
     begin
       fHelper.SetInputRange(APos - fHelperLen, APos, fInputContinue);
+      fHelper.ClearInternalExecData;
       if fHelper.MatchAtOnePos(APos - fHelperLen) then
       begin
         Result := False;
@@ -5621,6 +5623,10 @@ procedure TRegExpr.ClearMatches;
 begin
   FillChar(GrpBounds, SizeOf(GrpBounds), 0);
   FillChar(GrpSubCalled, SizeOf(GrpSubCalled), 0);
+end;
+
+procedure TRegExpr.ClearInternalExecData;
+begin
   FillChar(GrpSubCalled, SizeOf(GrpBacktrackingAsAtom), 0);
   IsBacktrackingGroupAsAtom := False;
 end;
@@ -5654,6 +5660,8 @@ begin
   // will lead to leaving ExecPrim without actual search. That is
   // important for ExecNext logic and so on.
   ClearMatches;
+  ClearInternalExecData;
+
 
   // Don't check IsProgrammOk here! it causes big slowdown in test_benchmark!
   if programm = nil then
@@ -5782,6 +5790,7 @@ end;
 
 procedure TRegExpr.SetInputRange(AStart, AEnd, AContinueAnchor: PRegExprChar);
 begin
+  ClearMatches;
   fInputString := '';
   fInputStart := AStart;
   fInputEnd := AEnd;
--- a/src/regexpr.pas
+++ b/src/regexpr.pas
@@ -4811,7 +4811,6 @@ var
   {$ENDIF}
   bound1, bound2: boolean;
   saveSubCalled: boolean;
-  savedIsBacktrackingGroupAsAtom: boolean;
 begin
   Result := False;
   {
@@ -5175,12 +5174,11 @@ begin
         begin
           no := Ord(scan^) - Ord(OP_OPEN);
           regCurrentGrp := no;
-//          savedIsBacktrackingGroupAsAtom := IsBacktrackingGroupAsAtom;
           save := GrpBounds[regRecursion].GrpStart[no];
           GrpBounds[regRecursion].GrpStart[no] := regInput;
           Result := MatchPrim(next);
           if GrpBacktrackingAsAtom[no] then
-            IsBacktrackingGroupAsAtom := False; // savedIsBacktrackingGroupAsAtom;
+            IsBacktrackingGroupAsAtom := False;
           GrpBacktrackingAsAtom[no] := False;
           if not Result then
             GrpBounds[regRecursion].GrpStart[no] := save;
--- a/src/regexpr.pas
+++ b/src/regexpr.pas
@@ -205,8 +205,8 @@ const
   RegExprUsePairedBreak: boolean = True;
   RegExprReplaceLineBreak: RegExprString = sLineBreak;
 
-  RegExprLookaheadIsAtomic: boolean = False;
-  RegExprLookbehindIsAtomic: boolean = True;
+  RegExprLookaheadIsAtomic: boolean = True;
+  RegExprLookbehindIsAtomic: boolean = False;
 
 const
   // Max number of groups.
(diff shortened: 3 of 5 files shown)


Alexey-T:
if #303 has #300 as its part, let's close #300 ?

Alexey-T:
`fInputContinue: PRegExprChar; // pointer to first char of input string` - misleading comment.

Alexey-T:
1. Please fill the history.txt, and adjust the version to 1.160 (in the unit + history.txt)
2. Pls test how the demo "test_dlg" shows opcode with \G

Alexey-T:
Thanks!

```
  Ptr := fInputStart + AOffset - 1;
  fInputContinue := Ptr;
```

if I got it, you set fInputContinue even on failing execs, but we must set it only on successfull exec.

User4martin:
Please check the history, if entries are ok.

Sorry about the comment, copy and paste is still a thing ;)

test_dlg now shows "Not fixed op: CONTINUE_POS"

Alexey-T:
Pls also comment 2 existing UniqueString calls.

User4martin:
> if https://github.com/andgineer/TRegExpr/pull/303 has https://github.com/andgineer/TRegExpr/pull/300 as its part, let's close https://github.com/andgineer/TRegExpr/pull/300 ?

If its fine to merge both in one, yes. (I.e. if you have no rule to merge those features separately)

Alexey-T:
Its fine to merge both. Andrey is busy man - he won't be happy to dig all this

Alexey-T:
Also: it's good to add \G to official docs, also in this repo.
pls add here:
https://regex.sorokin.engineer/en/latest/regular_expressions.html#boundaries

User4martin:
> Thanks!
> 
> ```
>   Ptr := fInputStart + AOffset - 1;
>   fInputContinue := Ptr;
> ```
> 
> if I got it, you set fInputContinue even on failing execs, but we must set it only on successfull exec.

This is set before the regex is run.

It is either the value given by the user `re.Exec(MyOffset)`

Or it is given by `ExecNext`, which already called `ExecPrim` with the offset from the last match (I have not changed that / the offset was stored on `GrpBounds[0].GrpEnd[0]` and will be stored into `fInputContinue` before we start the next match)

User4martin:
> Pls also comment 2 existing UniqueString calls.

Done in ad665552d93c00d4bdc8de0ba3843bc47760564d

Alexey-T:
`OP_CONTINUE_POS = Succ(OP_SUBCALL_LAST); ` won't work for undefined UnicodeRE because
```
  OP_SUBCALL_LAST =
    {$IFDEF UnicodeRE}
    TReOp(Ord(OP_SUBCALL) + RegexMaxGroups - 1);
    {$ELSE}
    High(REChar); // must fit to 0..255 range
    {$ENDIF}
```
so, insert new opcode in the middle, after the OP_RECUR. value 49 is good. seems OP_OPEN (=50) is not needing the change.

User4martin:
> Also: it's good to add \G to official docs, also in this repo. pls add here: https://regex.sorokin.engineer/en/latest/regular_expressions.html#boundaries

Done / not familiar with the markup used. Please check.

Also the section is called "Line Boundaries" and starts with"`.` any char" ? (anyway that is an issue of its own, if considered an issue)

Alexey-T:
oh, value 26 is good too!
```
 OP_EOL2 = TReOp(25); // like OP_EOL but also matches before final line-break
  OP_BSUBEXP = TREOp(28);
```

Alexey-T:
>Also the section is called "Line Boundaries" and starts with". any char" ? (anyway that is an issue of its own, if considered an issue )

right. please remove that line from the topic.

User4martin:
> ```
> `\G``        zero-length match at the end pos of the previous match. For ExecNext and Exec(Offset)
> ```
> 
> here is docs for usual users of programs, so the ending `. For ExecNext and Exec(Offset)` is not OK, remove it.

Ah, ok. I this case would this be good?

`\G``        zero-length match at the end pos of the previous match when using the g (global) modifier.

Alexey-T:
`when using the g (global) modifier.` is not OK enging- \G works always, we don't check 'g' modifier.

User4martin:
> `when using the g (global) modifier.` is not OK ending- \G works always, we don't check 'g' modifier.

Right, but when using the feature without writing the pascal code

Alexey-T:
> > `when using the g (global) modifier.` is not OK ending- \G works always, we don't check 'g' modifier.
> 
> Right, but when using the feature without writing the pascal code

does code check existing of 'g' modifier? not. so why to write this ending?

User4martin:
> > > `when using the g (global) modifier.` is not OK ending- \G works always, we don't check 'g' modifier.
> > 
> > 
> > Right, but when using the feature without writing the pascal code
> 
> does code check existing of 'g' modifier? no. so why to write this ending?

I hit the send button by accident.

Alexey-T:
parasitic dot is not  needed

re

Alexey-T:
you did good work! I don't have powers today to make all these testings in the test-project.

User4martin:
Btw, I did test the new StrLPos, but since it is hardly going to change, I did not write a test-case for it.

But for peace of mind, this is what I tested
```
a := 'abcdefaxdef';
writeln(StrLPos( PRegExprChar(a), PRegExprChar('a'), 0,1  ) - PRegExprChar(a) ); // to short
writeln(StrLPos( PRegExprChar(a), PRegExprChar('a'), 1,1  ) - PRegExprChar(a) ); //ok 
writeln(StrLPos( PRegExprChar(a), PRegExprChar('a'), 2,1  ) - PRegExprChar(a) ); // ok
writeln;
writeln(StrLPos( PRegExprChar(a), PRegExprChar('ab'), 0 ,2 ) - PRegExprChar(a) ); // to short
writeln(StrLPos( PRegExprChar(a), PRegExprChar('ab'), 1 ,2 ) - PRegExprChar(a) ); // to short
writeln(StrLPos( PRegExprChar(a), PRegExprChar('ab'), 2 ,2 ) - PRegExprChar(a) ); // ok
writeln(StrLPos( PRegExprChar(a), PRegExprChar('ab'), 7 ,2 ) - PRegExprChar(a) ); // ok
writeln;
writeln(StrLPos( PRegExprChar(a), PRegExprChar('ax1'), 0 ,2 ) - PRegExprChar(a) ); // to short
writeln(StrLPos( PRegExprChar(a), PRegExprChar('ax1'), 1 ,2 ) - PRegExprChar(a) ); // to short
writeln(StrLPos( PRegExprChar(a), PRegExprChar('ax1'), 2 ,2 ) - PRegExprChar(a) ); // to short
writeln(StrLPos( PRegExprChar(a), PRegExprChar('ax1'), 7 ,2 ) - PRegExprChar(a) ); // to short
writeln(StrLPos( PRegExprChar(a), PRegExprChar('ax1'), 8 ,2 ) - PRegExprChar(a) ); // ok
writeln(StrLPos( PRegExprChar(a), PRegExprChar('ax1'), 9 ,2 ) - PRegExprChar(a) ); //ok
writeln;
// all below do not find anything
writeln(StrLPos( PRegExprChar(a), PRegExprChar('fd'), 0 ,2 ) - PRegExprChar(a) );
writeln(StrLPos( PRegExprChar(a), PRegExprChar('fd'), 1 ,2 ) - PRegExprChar(a) );
writeln(StrLPos( PRegExprChar(a), PRegExprChar('fd'), 6 ,2 ) - PRegExprChar(a) );
writeln(StrLPos( PRegExprChar(a), PRegExprChar('fd'), 7 ,2 ) - PRegExprChar(a) );
writeln;
```

Alexey-T:
Can we have 'strLpos' name is good casing? and formatting of StrLScan with indent=2spaces?

Alexey-T:
Thanks

```
while (count < len) and (P[count] <> #0) do
```
StrLScan fails on chr0. But regex engine handles chr0 OK, so if you can improve it..

Alexey-T:
great!

Alexey-T:
lines are not needed in StrLPos
```
    if p = nil then
       exit;
```

Alexey-T:
you replaced
```
    if Pos(regMustString, fInputString) = 0 then Exit;
```
to
```
if StrLPos(fInputStart, PRegExprChar(regMustString), fInputEnd - fInputStart, length(regMustString)) <> nil then
        exit;
```

it's OK???

Alexey-T:
- FLAG_SPECSTART is set e.g. for this `\w*abcd`
- and ExecPrim must be run with SlowChecks=True, which is False by default

Alexey-T:
ShowChecks=False in some wrappers
```
function TRegExpr.Exec(const AInputString: RegExprString): boolean;
begin
  InputString := AInputString;
  Result := ExecPrim(1, False, False, False);
end; { of function TRegExpr.Exec
  -------------------------------------------------------------- }

{$IFDEF OverMeth}
function TRegExpr.Exec: boolean;
var
  SlowChecks: boolean;
begin
  SlowChecks := fInputEnd - fInputStart < fSlowChecksSizeMax;
  Result := ExecPrim(1, False, SlowChecks, False);
end; { of function TRegExpr.Exec
  -------------------------------------------------------------- }

function TRegExpr.Exec(AOffset: integer): boolean;
begin
  Result := ExecPrim(AOffset, False, False, False);
end; { of function TRegExpr.Exec
  -------------------------------------------------------------- }
{$ENDIF}   
```

Alexey-T:
Based on this info, can you please add testcase with regex `\w*abcd` and SlowChecks=True , testcase to run with regMustString.

Alexey-T:
Thank you

Alexey-T:
fInputContinue. when it is custom (not automatic)?
ExecPrim has
```
  Ptr := fInputStart + AOffset - 1;
  fInputContinue := Ptr;  
```

and all Exec* funcs call ExecPrim.
so, in all cases - fInputContinue is set to automatic value?
so, even if 
```
procedure TRegExpr.SetInputRange(AStart, AEnd, AContinueAnchor: PRegExprChar);
begin
  fInputString := '';
  fInputStart := AStart;
  fInputEnd := AEnd;
  fInputContinue := AContinueAnchor;
end;
```
is called, ExecPrim overrides custom fInputContinue value?

User4martin:
> fInputContinue. when it is custom (not automatic)? ExecPrim has
> ```
>   Ptr := fInputStart + AOffset - 1;
>   fInputContinue := Ptr;  
> ```
> 
> and all Exec* funcs call ExecPrim. so, in all cases - fInputContinue is set to automatic value? so, even if

"Custom" means `AOffset`. => `AOffset` can be given by the user (the person writing the Pascal code controlling the regex).

The user can give it via:
```
    function Exec(AOffset: integer): boolean; overload;
    function ExecPos(AOffset: integer {$IFDEF DefParam} = 1{$ENDIF}): boolean;
```

And for modifier "g", if you call `ExecNext` then `GrpBounds[0].GrpEnd[0]` is passed as `AOffset`.

`\G` is the position from which matching was continued (or started).
- If you start an entire new match, then it is 1 (AOffset will be passed with 1 as default)
- If you do ExecNext it is automatically done (the next match starts/continues at given pos, and \G is the anchor to that pos)
- If you use ExecPos or Exec(AOffset), then you start matching at AOffset, and the \G anchor is set accordingly.


> 
> ```
> procedure TRegExpr.SetInputRange(AStart, AEnd, AContinueAnchor: PRegExprChar);
> begin
>   fInputString := '';
>   fInputStart := AStart;
>   fInputEnd := AEnd;
>   fInputContinue := AContinueAnchor;
> end;
> ```
> 
> is called, ExecPrim overrides custom fInputContinue value?

`SetInputRange` is private, and is only called on fHelper used for look-behind. It follows the settings on the main instance.

Alexey-T:
```
procedure TRegExpr.SetInputRange(AStart, AEnd, AContinueAnchor: PRegExprChar);
begin
  fInputString := '';
  fInputStart := AStart;
  fInputEnd := AEnd;
  fInputContinue := AContinueAnchor;
end;
```
if we avoid setting fInputContinue here ^^^, what do we miss?

User4martin:
Normally if I continue to match (ExecNext / Exec(AOffset) the pattern can match at AOffset **or anywhere after AOffset**.

If you have a pattern, that you use to parse some text `\s*(var|const|type` then this can skip none matching content.
If you don't wont it to skip anything, you do `\G\s*(var|const|type`  ( `^` wont do, as it is not the start of line, but the start of the continuation match)

User4martin:
> if we avoid setting fInputContinue here ^^^, what do we miss?

`(?<!\G(?:var|const|type).*)\s+(.*)` match only if the continuation start pos did not begin with var/const or type.

Alexey-T:
yes, I got idea of these 2 regexes. but again, you didnot answer:

this line `fInputContinue := AContinueAnchor;` is ignored, because the ExecPrim (called in Exec*) will overwrite the fInputContinue, yes?

User4martin:
> yes, I got idea of these 2 regexes. but again, you didnot answer:
> 
> this line `fInputContinue := AContinueAnchor;` is ignored, because the ExecPrim (called in Exec*) will overwrite the fInputContinue, yes?

Ah, I see. I was wondering if I was missing something. The answer I gave was way to obvious.

So here is the answer.

No it does not overwrite it.

The only code calling it is
```
function TRegExpr.MatchAtOnePos(APos: PRegExprChar): boolean;
begin
...
      fHelper.SetInputRange(APos - fHelperLen, APos, fInputContinue);
      if fHelper.MatchAtOnePos(APos - fHelperLen) then
...

fHelper does not have a further helper, and goes to
```
function TRegExpr.MatchAtOnePos(APos: PRegExprChar): boolean;
begin
...
  Result := MatchPrim(regCodeWork);
```

`MatchPrim` skips `ExecPrim`

**Comment by Alexey-T** [comment]
now I see; thanks. I forgot about that fHelper code.

**Comment by Alexey-T** [comment]
@User4martin You are talent! thanks!

https://learn.microsoft.com/en-us/dotnet/standard/base-types/backtracking-in-regular-expressions
Let's rename IDs:

  FIsBactracingAtomic ->   GrpBacktrackingIsAtomic
  savedBackracingAtomic -> savedBacktrackingIsAtomic

**Comment by Alexey-T** [comment]
Increase the version to 1.161, and add History.txt info about this fix, please

**Comment by User4martin** [comment]
> Let's rename IDs:
> FIsBactracingAtomic -> GrpBacktrackingIsAtomic
> savedBackracingAtomic -> savedBacktrackingIsAtomic

Before I do, I think the new names are misleading at least... But mine aren't better.

This is not about the group being atomic (well by implication it is). But in an atomic group regular backtracking can happen before a first match is found.

What it should say is "The entire group is (to be) backtracked as one entity" => "The group is backtracked as a single atom"

So then it would be `IsBacktrackingGroupAsAtom` (with/without `F` not sure when that should be prefixed). 
(and `savedIsBacktrackingGroupAsAtom`)

**Comment by Alexey-T** [comment]
>So then it would be IsBacktrackingGroupAsAtom (with/without F not sure when that should be prefixed). (and savedIsBacktrackingGroupAsAtom)

Okay.

**Comment by Alexey-T** [comment]
Comment this var
`  savedIsBacktrackingGroupAsAtom: boolean;`

**Comment by Alexey-T** [comment]
Pls change in test_dlg/unit1.pas this
```
    if reg.IsFixedLength(op, i) then
      ListDump.Items.Add('Fixed length: '+IntToStr(i))
    else
      ListDump.Items.Add('Not fixed op: '+reg.DumpOp(op));
```
to this
```
    if reg.IsFixedLength(op, i) then
      ListDump.Items.Add('Match has fixed len: '+IntToStr(i))
    else
      ListDump.Items.Add('Match has not fixed len, operator: '+reg.DumpOp(op));
```

**Comment by Alexey-T** [comment]
New test for 'bads' now produces exception, it is OK, but better comment this test to avoid exceptions in test-app.

**Comment by User4martin** [comment]
> New test for 'bads' now produces exception, it is OK, but better comment this test to avoid exceptions in test-app.

Ehm, IMHO important to test that errors are detected.

How about: 
```
TTestRegExpr = class(TRegExpr)
  procedure Error(AErrorID: integer); override;
  property TestLastError: ....
end;
```

When called, store the error-id, and amend the test to check for that?

**Comment by Alexey-T** [comment]
Ok, leave this 'bad' test.

We must add clearing to ClearInternalIndexes
```
  FillChar(GrpBacktrackingAsAtom, SizeOf(GrpBacktrackingAsAtom), 0);
  IsBacktrackingGroupAsAtom := False;
```

**Comment by Alexey-T** [comment]
Or move clearing to ClearMatches (which is called much later)?

**Comment by User4martin** [comment]
> Or move clearing to ClearMatches (which is called much later)?

ClearMatches seems the better choice to me.

The data isn't set by compiling, it is used in each run of the reg-ex. So if the previous run did not exit clear, then it must be cleared before the next run.

I will add it to ClearMatches.

**Comment by User4martin** [comment]
> Ok, leave this 'bad' test.

Would it be ok to add a subclass and prevent the exception?

Or otherwise, add ERegExpr as "ignored exception class" to the testcase?

**Comment by User4martin** [comment]
Maybe rename `ClearMatches` to `ClearBeforeExec` ?

**Comment by Alexey-T** [comment]
>Would it be ok to add a subclass and prevent the exception?

yes, please.

>Maybe rename ClearMatches to ClearBeforeExec ?

yes.

**Comment by User4martin** [comment]
I found a missing call to ClearMatches.

I separated ClearMatches and ClearInternalExecData => Matches get cleared on various location. Internal data only matters before the Exec.

Added the error testing.
Outside of expected error testing, it will still raise (and alert to potential issues). While expecting an error, it will prevent the exception (knowing that the result will be checked)
(Enabled the pre-existing error tests / since there no longer is a reason for them to be commented)

**Comment by Alexey-T** [comment]
` FillChar(GrpSubCalled, SizeOf(GrpBacktrackingAsAtom), 0);` - why 2 different names?

**Comment by Alexey-T** [comment]
I feel that ClearMatches must be called always together with ClearInternalExecData (so it must be a single procedure)

**Comment by User4martin** [comment]
> ` FillChar(GrpSubCalled, SizeOf(GrpBacktrackingAsAtom), 0);` - why 2 different names?

ouch...

> I feel that ClearMatches must be called always together with ClearInternalExecData (so it must be a single procedure)

`ClearMatches` is called when changing the `InputString` => that is needed because the PChar no longer point to something valid.

If the user changes the `InputString` 10 times (without calling `Exec`), then the `GrpBacktrackingAsAtom` does not need to be cleared each time. It isn't going to be accessed until the next `Exec`

So technically they want to be different.

Practically, it makes little difference. It isn't much extra time spent. (And the scenario is unlikely...)
Well, for the common sequence
- `re.InputString := foo;`
- `re.Exec`

`ClearMatches` is already called twice. If a user loops the above 1000 times, then it is extra processing (percentage is still insignificant).
Keeping `ClearInternalExecData` will at least keep that to be run only once per exec.

**Comment by User4martin** [comment]
Any idea which commit changed that?

**Comment by User4martin** [comment]
> Thanks. The remaining problem (regression) is the `Inc(LoopStackIdx)`/`Dec()`. `Dec` is called too much. it was OK in the 'release' version.

Is that tag 1.155 ?

I tested, I get the range check for the first 2 examples from the issue #307.
The last example (with the `{}`) fails to match (even though it should match)

**Comment by Alexey-T** [comment]
I mean version 1.159 (last Git before you make changes): first example from #307 don't give range error/

**Comment by User4martin** [comment]
> I mean version 1.158 (last Git before you make changes): first example from #307 don't give range error.

revision 39aaae8d37d3399a04f84c14cd8b6c0d38abfdf6

Both below fail with the range check

```
var
  RegEx: TRegExpr;

begin
RegEx := TRegExpr.Create('A(?>(?:b|c(?:d|e)*?)*?_)a');
RegEx.InputString := 'Acece_x_Acece_a_';
RegEx.Exec;

RegEx := TRegExpr.Create('A(?>(?:b|c(?:d|e(?:f|g)*?)*?)*?_)a');
RegEx.InputString := 'Acece_x_Acecegg_a_';
RegEx.Exec;
```

**Comment by Alexey-T** [comment]
(A guess: some var/array is not cleared between different Match* calls?)

**Comment by Alexey-T** [comment]
Pls remove 'inline' from `MatchAtOnePos` - it is big function and it's called 3 times.

**Comment by User4martin** [comment]
> Pls remove 'inline' from `MatchAtOnePos` - it is big function and it's called 3 times.
> 
> If Inc/Dec range-error is not OUR regression, then it's ok to apply PR.

I didn't add it. But will do... Would do... See below

Btw, I only see 2 calls. One is in a loop.

I am working on some further stuff (more soon). And that functions will reduce to less than 10 lines (half of them simple assignment, that produce little asm...)

I feel the small version, would be ok.

Just to say, that "some stuff" makes lots of changes => so it be safer in its own PR.

**Comment by Alexey-T** [comment]
So is this PR ready?

**Comment by User4martin** [comment]
> So is this PR ready?

If you haven't found any other issue with it, then yes.

**Comment by Alexey-T** [comment]
@andgineer Please apply?

**Comment by Alexey-T** [comment]
The compat with old Delphi needs this define, which must disable 'overload' methods
```
    constructor Create; {$IFDEF OverMeth} overload;
    constructor Create(const AExpression: RegExprString); overload;
    {$ENDIF}
```
so IsFixedLength overloads should be renamed

User4martin:
It seems you added and deleted a comment on "overload" => can be ignored ?

Alexey-T:
I moved the comment to another commit.

User4martin:
> I moved the comment to another commit.

Ok... I didn't get a notification on the new location. Hence to me it completely disappeared.

But I take it the answer the is "no - not ignored" since it (the comment) still is somewhere.

Will look into, renaming them

User4martin:
Will the name `IsPartFixedLength` be ok?
`    function IsPartFixedLength(var prog: PRegExprChar; var op: TREOp; var ALen: integer; StopAt: TREOp): boolean; overload;`

I.e.: the part from `prog` to `StopAt`

Alexey-T:
>Will the name IsPartFixedLength be ok?

yes, for new overloads. Old name for public method must not change.

Issue #307, opened 2023-08-17
Range check error in OP_LOOP, OP_LOOPNG:


User4martin:
regex `A(?>(?:b|c(?:d|e)*?)*?_)a`
text `Acece_x_Acece_a_`

produce a range check error
`LoopStackIdx` = 0  for `TRegExprLoopStack = array [1 .. LoopStackMax] of integer;`

revision 39aaae8d37d3399a04f84c14cd8b6c0d38abfdf6

Alexey-T:
w/out thinking I changed lower index of this type to 0
```
type
  TRegExprLoopStack = array [0 .. LoopStackMax] of integer;     
```
now demo test_dlg handles it

User4martin:
Looking at
```
    LoopStackIdx: integer; // 0 - out of all loops
```

I don't think that is correct.

It gets increased in 
```
      OP_LOOPENTRY:
        begin // ###0.925
          no := LoopStackIdx;
          Inc(LoopStackIdx);
```

OP_LOOP, OP_LOOPNG: increases and decreases it.  

But that is as far as I looked.

User4martin:
But just to prove the point: I changed it to start at 0, and then run the below.

RegEx `A(?>(?:b|c(?:d|e(?:f|g)*?)*?)*?_)a`
Text: `Acece_x_Acecegg_a_`

The index goes to -1

User4martin:
Interesting is, that it does not crash, if the text has a single "g", only with the 2nd "g".

Maybe that holds some clue. But currently got other stuff to do.

Alexey-T:
I set breakpoint on  Dec(LoopStackIdx) and see that this Dec called to much, in the handled of 
`OP_LOOP, OP_LOOPNG: `

Alexey-T:
Simple fix: after
`Dec(LoopStackIdx);`
add
```
            if LoopStackIdx < 1 then
                LoopStackIdx := 1; 
```

User4martin:
Then it does not match 
https://regex101.com/r/7FAwus/1

`A(?>(?:b|c(?:d|e(?:f|g){0,2}?){1,1}?){2,2}?_)a`
`Acece_x_Acecegg_a_`

Not sure if that is the same bug, but I thing so because
`A(?>(?:b|c(?:d|e(?:f|g){0,2}?){0,1}?){2,2}?_)a`
does match.

I only changed `{1,1}` to `{0,1}`

[...]

PR #311, opened 2023-08-19
Fix loop index #307

User4martin:
Fix for issue #307
See issue for description/reason of changes.

Missing entries in history.txt => this depend on order of merging compared with other merge requests.

Incudes unrelated addition, to allow `RE.Dump` display loops and open/close with indent.


Alexey-T:
very good. @andgineer Please apply only after \G pull-request, it is the main.

User4martin:
Please HOLD

While this fixes some issues, it is not correct. And I found another pattern to break it.

User4martin:
replaced by #312

Issue #304, opened 2023-08-15
Documentation has some items under wrong heading


User4martin:
```
Boundaries
----------

.. _lineseparators:

Line Boundaries
~~~~~~~~~~~~~~~

============= ================================================
Meta-char     Matches
============= ================================================
``.``         any character, can include line-breaks
```

Not a line boundary. Should be moved.

User4martin:
I can see why it may have been put there: It can match a line-break.

But a line-break is not itself the line boundary. The boundary is a zero length construct (BOL or EOL).

Alexey-T:
@User4martin I posted the PR, ok?

PR #318, opened 2023-09-02
docs, fix https://github.com/andgineer/TRegExpr/issues/304

Alexey-T:


User4martin:
Further down in the "boundary" part 
```
Meta-char ``.`` (dot) by default matches any character, but if you
turn **off** the `modifier /s <#s>`_, then it won't match line-breaks inside the string.
```

Also needs to move.

Alexey-T:
Thanks for note, moved now.

User4martin:
Better.... The more I look, the more I find.

The meta-classes in the table below have the note 
```
You may use all meta-classes, mentioned in the table above, within
`user character classes <User Character Classes_>`_.
```
Which does not extend to the "dot", but may be misread.

----

I would
- move the "dot meta" below the "meta classes table"
- Add to the "dot meta"
```
The ``.`` does not act as meta class inside `user character classes <User Character Classes_>`_.
``[.]`` means a literal "."
```

Alexey-T:
'Dot' topic is now separate:

scr

User4martin:
looks ok to me

PR #320, opened 2023-09-03
Clear last error, before execution.

### LINKED ISSUE #310: LastError not always cleared
The first call to `Exec` fails and sets LastError.
The 2nd call however succeeds. Yet LastError is still set.

On the other hand, if one has an error compiling a RegEx (`r.Expression := ...; r.Compile`) then LastError is cleared (in case a prior compile or exec had set an error. Actually 
Clearing happens in `CompileRegExpr`, it might want to be done earlier (e.g. when the new Expression is assigned)

```
var
  RegEx: TRegExpr;

begin
  RegEx := TRegExpr.Create('a');
  RegEx.InputString := 'foobar';
  try
    writeln(1);
    writeln(RegEx.Exec(-9));
  except
  end;
  try
    writeln(2);
    writeln(RegEx.Exec(2));
  except
  end;
  writeln(RegEx.LastError);
```

output:
```
1
2
TRUE
1006
```

User4martin:
fix #310

### DIFF
--- a/src/regexpr.pas
+++ b/src/regexpr.pas
@@ -5963,6 +5963,7 @@ end;
 
 procedure TRegExpr.ClearInternalExecData;
 begin
+  fLastError := reeOk;
   FillChar(GrpBacktrackingAsAtom, SizeOf(GrpBacktrackingAsAtom), 0);
   IsBacktrackingGroupAsAtom := False;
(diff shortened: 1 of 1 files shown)


Alexey-T:
@andgineer Please apply...

PR #322, opened 2023-09-10
Fix SetInputSubString for zero-length input

User4martin:
SetInputSubString  did not set the input to empty, if it was given an empty string, or len=0


User4martin:
@Alexey-T

Alexey-T:
I looked at the patch, seems valid.

User4martin:
now part of #325

PR #323, opened 2023-09-11
Add "function Exec(AOffset, AMatchMustStartBefore: integer): boolean;"

User4martin:
@Alexey-T 

Would this be of interest?

The use case for it is, when searching through a list of patterns for the one that matches a given text first
```
  for regx in list_of_many_reg_ex do
    if matchpos(regx, text) < known_first then
      known_first := ...
```

If any regex matched at pos 20, then the rest of the list does not need to search for matches after that pos.

### DIFF
--- a/src/regexpr.pas
+++ b/src/regexpr.pas
@@ -556,7 +556,7 @@ type
     function MatchAtOnePos(APos: PRegExprChar): boolean; {$IFDEF InlineFuncs}inline;{$ENDIF}
 
     // Exec for stored InputString
-    function ExecPrim(AOffset: integer; ATryOnce, ASlowChecks, ABackward: boolean): boolean;
+    function ExecPrim(AOffset: integer; ATryOnce, ASlowChecks, ABackward: boolean; ATryMatchOnlyStartingBefore: Integer): boolean;
 
     function GetSubExprCount: integer;
     function GetMatchPos(Idx: integer): PtrInt;
@@ -612,7 +612,12 @@ type
     // (AOffset=1 - first char of InputString)
     function ExecPos(AOffset: integer {$IFDEF DefParam} = 1{$ENDIF}): boolean;
     {$IFDEF OverMeth} overload;
+    // find match for InputString at AOffset.
+    // if ATryOnce=True then only match exactly at AOffset (like anchor \G)
+    // if ATryMatchOnlyStartingBefore then only when the match can start before
+    // that position: Result := MatchPos[0] < ATryMatchOnlyStartingBefore;
     function ExecPos(AOffset: integer; ATryOnce, ABackward: boolean): boolean; overload;
+    function ExecPos(AOffset, ATryMatchOnlyStartingBefore: integer): boolean; overload;
     {$ENDIF}
 
     // Returns ATemplate with '$&' or '$0' replaced by whole r.e.
@@ -5906,7 +5911,7 @@ end; { of function TRegExpr.MatchPrim
 function TRegExpr.Exec(const AInputString: RegExprString): boolean;
 begin
   InputString := AInputString;
-  Result := ExecPrim(1, False, False, False);
+  Result := ExecPrim(1, False, False, False, 0);
 end; { of function TRegExpr.Exec
   -------------------------------------------------------------- }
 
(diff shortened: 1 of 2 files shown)


Alexey-T:
We already have
`    function ExecPrim(AOffset: integer; ATryOnce, ASlowChecks, ABackward: boolean): boolean;`

for this purpose. set `ATryOnce=True`.

Alexey-T:
I mean - call ExecPos(... ATryOnce=True ...) and adjust the AOffset manually before each call. If you have spaces/tabs - your code can just skip them. That is what 'lite lexer' in CudaText does.

User4martin:
> I mean - call ExecPos(... ATryOnce=True ...) and adjust the AOffset manually before each call. If you have spaces/tabs - your code can just skip them w/o calling ExecPos. That is what 'lite lexer' in CudaText does.

But would that not loose the
```
      if not FirstCharArray[byte(Ptr^)] then
```
optimization?

User4martin:
Well, yes => for my case I run in utf8 (ansichar) mode. => because the syn-highlighter gets text in utf8.

But Ok, I expected this may not be for the official release. I will probably end up running a copy of the code for what I need. So I can still use this change.

Feel free to close the PR as "not merged".

User4martin:
Ups, yes.

But, ups there is an other issue.

I can't just pass in different offsets. Because the offset affects \G.

Which may be easy to check, if it is at the very start of the regex, but what if
`abc|\G123`
`foobar.something(?<=\G123)`

If I pass in a higher offset, to search at subsequent positions, then those patterns match the \G wrong.

Alexey-T:
I don't know how to solve this yet.

Alexey-T:
`abc|\G123` - why it will not work? \G will be checked at the AOffset of Exec.

User4martin:
> `abc|\G123` - why it will not work? \G will be checked at the AOffset of ExecPos. or not?

Well I have a text `CurrentLine". Let's say I have matched the first 10 chars already, and now want to find the next pattern, that matches from pos=11.

I would do
```
for i := 0 to length(regex)-1 do begin
  if regex[i].Exec(11) then 
    MatchPos := regex[i].MatchPos[0];
  LowestMatchPos := min(LowestMatchPos, MatchPos);
end;
```

The actually, instead of `regex[i].MatchPos[0]`, I only want the match, if it starts before `LowestMatchPos`.

But if I do
```
  for x := 11 to LowestMatchPos -1 do 
     regex[i].ExecPos(x, True) // True = OnlyOnce
```

Then \G will could at each of the `x` given. Yet it should only match at 11.

So I can't give any higher x to `Exec/ExecPos`.

Alexey-T:
Maybe we need additional param like AUseOffsetForGAnchor ?

User4martin:
If we add an extra param, then why not the one I added? (proposed to add).
There may be a better name than `AMatchMustStartBefore`.

But at least that way it uses the existing loop (inside ExecPrim), and user-code does not need to write a 2nd loop on the outside.

Also, in future there may be other optimizations inside that loop in ExecPrim.
Actually there are already
`.*abc` sets the raOnlyOnce flag. So if I want to know if `.*abc` matches my text, but only if it starts before `AMatchMustStartBefore` then there will be no loop at all. If it doesn't match directly at `AOffset`, then it wont match anywhere.
But if I have my own loop, then my own loop will try it in all the possible subsequent positions....

Alexey-T:
okay. It may be like you suggested. 

- Why do you name new overload Exec, not ExecPos?
- pls use `{$IFDEF DefParam}` and `{$IFDEF OverMeth}`

User4martin:
> okay. It may be like you suggested.
> 
>     * Why do you name new overload Exec, not ExecPos?

To me, one was as good as the other. 
- Both have the `(AOffset)`
- Both need an overload

Well, ok it is somehow related to `ATryOnce` / It is a more generic `ATryOnce`.

Will rename the function.

That might also be an idea for the param name?
Despite `ATryOnce` itself being rather non-descriptive...
For the new value: `ATryMatchOnlyStartingBefore`


> 
>     * pls use `{$IFDEF DefParam}` and `{$IFDEF OverMeth}`

The latter is already used.

For `ExecPrim` => I had not seen (paid attention) that def-param are restricted. 
I will remove the default for ExecPrim, and amend all callers to supply a value.

Alexey-T:
>For the new value: ATryMatchOnlyStartingBefore

Ok.

Alexey-T:
Looks good. I must check it in my text editor with 'lite'l exers, todo.

Alexey-T:
'Lite' lexers uses different regex engine, sorry. Ok, if this is tested good, let's apply.

User4martin:
now part on #325

PR #325, opened 2023-09-12
Fix recursion

### LINKED ISSUE #324: Recursion and backreferences
This may not be a bug, but at least a question how it should be expected to work.

https://regex101.com/r/FrdqmJ/1

`(?:x|([abc]))(?R)?-\1*`
matches `aabxa-a-b-b-a-a`

Same match with `(?:x|([abc]))(?R)?-\1` (no `*` for the backref)

From most inner recursion to outwards
- `a-a`  Captures `a` and `\1` matches the `a`
- `x___-b`  does not capture but `\1` matches the `b` of the calling (next outer) recursion.
- `b_____-b` Captures `b` and `\1` matches the `b`
- .....

However TRegExpr

`(?:x|([abc]))(?R)?-\1*` matches `xa-a-`
`(?:x|([abc]))(?R)?-\1` matches `a-a`

changing the text to `aabxa-a--b-a-a`
`(?:x|([abc]))(?R)?-\1*` matches the full text

In conclusion: With TRegExpr the  backreference `\1` - if the current recursion has no capture of its own - does not match the outer callers capture.

That may be seen as valid or not. It is at first a decision, and then possibly a documentation issue.

----
It may also want to be documented, that if the capture for a back reference does either not exist, or has not been triggered yet, then in TRegExpr the back-ref returns false.

This is the same in PCRE.
But the ECMA engine  matches empty in that case.
https://regex101.com/r/7C9IM7/1

User4martin:
Fix #324 comment 1. Subsequent (side-by-side) recursive calls did see data captured in previous calls.
Improve speed.

Include existing PR #323  and #322 

@Alexey-T 

All tests work for the old and new recursion behaviour. Except
```
IsMatching('2nd recurse does NOT see capture from earlier recurse',
             '[aA](?R)?(?:X|([bcBC]))(?R)?\1',
             'aABBcAXBc',  [2,3,  3,1]);

  IsMatching('2nd recurse does NOT see capture from earlier recurse',
             '[aA]((?R))?(?:X|([bcBC]))((?R))?\2',
             'aABBcAXBc',  [2,3,  -1,-1,  3,1,  -1,-1]);
```

Which are specially crafted to exploit the issue in the old behaviour

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -6,11 +6,16 @@
                                      (-) - bugfix
                                      (^) - improvement
 
- v. 1.166 2023.09.13
+ v. 1.167 2023.09.13
  -=- (+) Added Exec(AOffset, AMatchMustStartBefore: integer)
   Limit the search for the first matched position to AMatchMustStartBefore
+ -=- (+) Added named group syntax: (?<name>regex)
+ -=- (+) Added named group reference/call syntax with \g and \k
+ -=- (^) Allow underscore in named group name
  -=- (-) Fixed recursive calls accessing outer captures
  -=- (^) Improved speed by reducing clean-up before each match execution
+ -=- (-) Fixed literal brace "{" after quantifier.
+  By Martin Friebe.
 
  v. 1.165 2023.09.08
  -=- (+) Added support for \R (any unicode line break).
--- a/History.txt
+++ b/History.txt
@@ -6,6 +6,12 @@
                                      (-) - bugfix
                                      (^) - improvement
 
+ v. 1.166 2023.09.13
+ -=- (+) Added Exec(AOffset, AMatchMustStartBefore: integer)
+  Limit the search for the first matched position to AMatchMustStartBefore
+ -=- (-) Fixed recursive calls accessing outer captures
+ -=- (^) Improved speed by reducing clean-up before each match execution
+
  v. 1.165 2023.09.08
  -=- (+) Added support for \R (any unicode line break).
   by Alexey Torgashin.
--- a/docs/regular_expressions.rst
+++ b/docs/regular_expressions.rst
@@ -428,14 +428,18 @@ The entire regex has index 0.
 Backreferences
 --------------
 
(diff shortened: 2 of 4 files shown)


Alexey-T:
@andgineer Pls apply.

Alexey-T:
(Maybe 'VersionMinor' change is conflict?)

Alexey-T:
>Can that be added?
https://regex101.com/r/7qgquy/1

It must be discussed in a new topic. with detailed description: what was not ok, what you propose...

User4martin:
Added #326
Added fix for "literal {" mode

Alexey-T:
Write to history block ending: "by Martin Friebe."
Better increase version to 1.167.

User4martin:
Added another fix. Checking fixed length of a look-behind would fail if the look-behind had certain reference or sub-calls. (Because the HasFixedLength would increment the pointer to the next OP, but not skip the index stored after the command.)

Alexey-T:
```
``\k`name```
```
Did you make mistake - backtick instead of single quote?

User4martin:
@Alexey-T 

- Added: "GrpEnd" is guarded by "GrpStart<>nil" => no need to "FillChar" (speed up)
- fix: An earlier (already merged) commit added "default param"
- `GrpSubCalled` is only needed in execution loop, and therefore should only be cleared for a new Exec call

Alexey-T:
Good, tkx.

```
 (GrpBounds[0].GrpEnd[Idx] > GrpBounds[0].GrpStart[Idx])
```

maybe better `>=` ?

User4martin:
> Good, tkx.
> 
> ```
>  (GrpBounds[0].GrpEnd[Idx] > GrpBounds[0].GrpStart[Idx])
> ```
> 
> maybe better `>=` ?

That is in `TRegExpr.GetMatch` ?

If it is equal, then the result is an empty string. And it has already been set to empty.

Alexey-T:
Ok, indeed.

Alexey-T:
@andgineer Please apply?

PR #328, opened 2023-09-15
Review / Variable amount of Groups

### LINKED ISSUE #327: Inconsistent substitute $123
The below code basically means, that `NumberFound` will be true for any `$1` to `$89`
But not for `$90` and up.

That seems are strange choice.

```
  RegexMaxGroups = 90;
    GrpIndexes: array [0 .. RegexMaxGroups - 1] of integer; // map global group index to _capturing_ group index
```

```
  procedure FindSubstGroupIndex(var p: PRegExprChar; var Idx: integer; var NumberFound: boolean);
  begin
    Idx := ParseVarName(p);
    NumberFound := (Idx >= 0) and (Idx <= High(GrpIndexes));
```

The test asserts, that a number not present is ignored. But that works only up to 89.
And then behaviour changes. 

```
    // 70
    (
    expression: '(b)';
    inputText: 'abc';
    substitutionText: '$1$2$9';
    expectedResult: 'abc';
    matchStart: 1
    ),
```

Changing that test to `$1$2$90`  leads to a result of `ab$c`

IMHO, any $123 ... that does not have a group should be "".

User4martin:
This
- fixes #327 
- allows (P=name) to look forward (like \1 can refer forward)
- Allows "unlimited" groups (well the list must fit into memory) :)
- removes some more FillChar

A count of groups is taken during the first pass. Then all arrays are set to their required length.

Only exception is the name array. As it needs to be populated during the first pass.
It only is grown, if an actual named-capture is found, and a name needs to be stored. 

@Alexey-T

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -6,6 +6,11 @@
                                      (-) - bugfix
                                      (^) - improvement
 
+ v. 1.169 2023.09.16
+ -=- (^) Removed limit to "Max number of groups" (RegexMaxGroups).
+ -=- (-) Fixed recursive sub calls. Sub call could return to early.
+  By Martin Friebe.
+
  v. 1.167 2023.09.13
  -=- (+) Added Exec(AOffset, AMatchMustStartBefore: integer)
   Limit the search for the first matched position to AMatchMustStartBefore
--- a/src/regexpr.pas
+++ b/src/regexpr.pas
@@ -123,7 +123,7 @@ uses
     Character,
     {$ENDIF}
   {$ENDIF}
-  Classes; // TStrings in Split method
+  Classes, Math; // TStrings in Split method
 
 type
   {$IFNDEF FPC}
@@ -210,11 +210,11 @@ const
   // Be carefull - don't use values which overflow OP_CLOSE* opcode
   // (in this case you'll get compiler error).
   // Big value causes slower work and more stack required.
-  RegexMaxGroups = 90;
+  RegexGroupCountIncrement = 50;
 
-  // Max possible value for RegexMaxGroups.
+  // Max possible amount of groups.
   // Don't change it! It's defined by internal TRegExpr design.
-  RegexMaxMaxGroups = 255;
+  RegexMaxMaxGroups = MaxInt div 16;
 
   // Max depth of recursion for (?R) and (?1)..(?9)
   RegexMaxRecursion = 20;
@@ -282,8 +282,8 @@ type
 
 
   TRegExprBounds = record
-    GrpStart: array [0 .. RegexMaxGroups - 1] of PRegExprChar; // pointer to group start in InputString
-    GrpEnd: array [0 .. RegexMaxGroups - 1] of PRegExprChar; // pointer to group end in InputString
+    GrpStart: array of PRegExprChar; // pointer to group start in InputString
+    GrpEnd: array of PRegExprChar; // pointer to group end in InputString
   end;
   TRegExprBoundsArray = array[0 .. RegexMaxRecursion] of TRegExprBounds;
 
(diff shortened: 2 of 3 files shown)


Alexey-T:
I wrote new issue #330, other code seems good.

Alexey-T:
outdated comments, maybe clear them?
```
  // Max number of groups.
  // Be carefull - don't use values which overflow OP_CLOSE* opcode
  // (in this case you'll get compiler error).
  // Big value causes slower work and more stack required.
  RegexGroupCountIncrement = 50;

  // Max possible amount of groups.
  // Don't change it! It's defined by internal TRegExpr design.
  RegexMaxMaxGroups = MaxInt div 16;
```

Alexey-T:
Unclear code for me
```

function GroupDataArraySize(ARequired, ACurrent: Integer): Integer;
begin
  Result := ARequired;
  if Result > ACurrent then
    Exit;

  // Keep some extra
  if Result > ACurrent - RegexGroupCountIncrement then
    Result := ACurrent;
end;
```

User4martin:
When resizing those arrays, and the new value is less than the existing, then some space is kept.

This only happens if the expression is changed (new expression compiled). If a user changes the expression once, then likely it will change again, and later and require the memory again.

Alexey-T:
Okay then.
Reviewed, seems good.

User4martin:
@Alexey-T 

final review please.

Alexey-T:
Seems good. @andgineer

Alexey-T:
pls, some overflow check is needed, when we calculate index of grp: Index:= Index*10+Digit.

Alexey-T:
typo: `out ANumpber`

andgineer:
Are we good to merge?

User4martin:
@Alexey-T have you seen the latest amends? Ok?

Alexey-T:
Ok from me.

Alexey-T:
```
  SetLength(GrpIndexes, GroupDataArraySize(regNumBrackets, Length(GrpIndexes)));
  for i := 1 to regNumBrackets - 1 do
    GrpIndexes[i] := -1;
  GrpIndexes[0] := 0;

  SetLength(GrpOpCodes, GroupDataArraySize(regNumBrackets, Length(GrpOpCodes)));
  SetLength(GrpBacktrackingAsAtom, GroupDataArraySize(regNumBrackets, Length(GrpBacktrackingAsAtom)));
```
so we now don't clear these arrays. we set some length considering previous length. 
so we can keep part of old data, from previous RegEx compilation?
e.g. old RegEx was complex, new RegEx is simple.

Alexey-T:
Does code read that old items in these arrays, or it reads onky first regNumBrackets items which are valid for current regex?

Alexey-T:
2. Do we need to free all arrays in Destroy or compiler does it?

User4martin:
Dynamic arrays are refcounted => they get freed.

The values get cleared during compilation.

GrpIndexes:
There is only one place where ParsedGrpCount gets incremented, and GrpIndexes is set for each value
```
                Inc(ParsedGrpCount);
                if fSecondPass then begin
                  GrpIndexes[ParsedGrpCount] := regNumBrackets;
```

GrpNames:
Is now managed in its own object, and count is only increased for values added (on gaps)


GrpOpCodes also set for each index
```
      Inc(regNumBrackets);
      if BeginGroupOp <> OP_NONE then
        ret := EmitNodeWithGroupIndex(BeginGroupOp, NBrackets);
      if fSecondPass then
        GrpOpCodes[NBrackets] := ret;
```

the others are cleared before execution.

-----
Just seeing.
`GrpIndexes`
are over allocated (always have been).

They only need GrpCount as length => The only exist for capturing groups, and not for others.

Alexey-T:
Ok to merge.

PR #347, opened 2023-11-17
Optimize branch

User4martin:
The last section of a branch `(abc|def|last)` does not need a recursive call. Minor speed up.


Alexey-T:
@User4martin You need 'pull' as I removed old comments.

PR #354, opened 2023-11-18
Optimize branch 

User4martin:
Optimize `|` branches

Fix for bug #352 

```

                                                        Time     Before | Opt-Branch     | Match count
========================================================================================================
                                                  /Twain/ :       41 ms |      37 ms     |      2388
                                              /(?i)Twain/ :       58 ms |      58 ms     |      2657
                                             /[a-z]shing/ :      228 ms |     228 ms     |      1877
                             /Huck[a-zA-Z]+|Saw[a-zA-Z]+/ :       39 ms |      35 ms     |       396
                                                      /./ :      421 ms |     449 ms     |  18905427
                                                    /(.)/ :      660 ms |     707 ms     |  18905427
                                                      /e/ :       84 ms |      80 ms     |   1781425
                                                    /(e)/ :      105 ms |     103 ms     |   1781425
                                           /(?s).{1,45} / :       37 ms |      37 ms     |    475715
                                         /(?s)\G.{1,45} / :      164 ms |     144 ms     |     10616
                                         /\G(?s).{1,45} / :       15 ms |      15 ms     |     10616
                                          /(?s).{1,45}? / :      154 ms |     160 ms     |   3241534
                                        /(?s)\G.{1,45}? / :      169 ms |     146 ms     |     69431
                                        /\G(?s).{1,45}? / :       19 ms |      19 ms     |     69431
                                              /\b\w+nn\b/ :      230 ms |     226 ms     |       359
                                       /[a-q][^u-z]{13}x/ :      652 ms |     613 ms     |      4929
                            /Tom|Sawyer|Huckleberry|Finn/ :       42 ms |      39 ms     |      3015
                        /(?i)Tom|Sawyer|Huckleberry|Finn/ :      164 ms |     142 ms +   |      4820
                    /.{0,2}(Tom|Sawyer|Huckleberry|Finn)/ :     2123 ms |    1636 ms +++ |      3015
                    /.{2,4}(Tom|Sawyer|Huckleberry|Finn)/ :     2330 ms |    1820 ms +++ |      2220
                      /Tom.{10,25}river|river.{10,25}Tom/ :       56 ms |      50 ms     |         2
                                           /[a-zA-Z]+ing/ :      585 ms |     574 ms     |     95863
                                  /\s[a-zA-Z]{0,12}ing\s/ :      261 ms |     257 ms     |     67810
                          /([A-Za-z]awyer|[A-Za-z]inn)\s/ :      527 ms |     521 ms     |       313
[...]

PR #355, opened 2023-11-18
Fix look around

User4martin:
There was a bug in lookaround patterns. See commit description.

### DIFF
--- a/src/regexpr.pas
+++ b/src/regexpr.pas
@@ -524,6 +524,9 @@ type
     // regnode - emit a node, return location
     function EmitNode(op: TREOp): PRegExprChar;
 
+    // emit OP_BRANCH (and fillchars)
+    function EmitBranch: PRegExprChar;
+
     // emit (if appropriate) a byte of code
     procedure EmitC(ch: REChar); {$IFDEF InlineFuncs}inline;{$ENDIF}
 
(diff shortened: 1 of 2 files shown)


Alexey-T:
i applied the #354, and it has the common 2 commits with this PR. how should I apply this new PR?

Alexey-T:
This RE looks strange, from tests

`'^.(?:a|.(?=2)?|b)'`

quantifier ? stands after the look-around. is it correct?

User4martin:
> `'^.(?:a|.(?=2)?|b)'`
> 
> quantifier ? stands after the look-around. is it correct? @User4martin

Yes. A look around may be optional. 

The page https://regex101.com/r/fDHYug/1 says
> A quantifier following a lookaround serves no purpose, and can safely be removed from the regular expression

But that is not correct `1(?=(2))?` will match a "1" regardless of it being followed by a "2" or not. But the capture group will be set.
https://regex101.com/r/5x6kzc/1

Note that `1(2)?` is different, because then the entire match (on the text "12") becomes "12" instead of "1" with group[1]="2"

However, I think it can be rewritten as `1(?=(?:(2))?)` making the entire content of the lookahead optional.

Alexey-T:
Thanks for this info.

User4martin:
Just realized, still need the possessive version: https://regex101.com/r/g5UMHG/1 `^(.*?)1(?=(2))?+.*x1(?!\2)`  remove the `+` and it changes.

PR #356, opened 2023-11-19
Fix overflow in IsPartFixedLength / remove incorrect assert.

User4martin:
2 small fixes / see commit description


User4martin:
The assert is wrong. In a branch the next pointer does point to the end of the branch. 

However the code is otherwise correct. I.e. it should do `s:=next`.
The assert was merely testing a temp value that was never used.

PR #365, opened 2023-11-24
Use gbranch in *+ and ?

User4martin:
Since loops use branches (in some cases) let they benefit from the OP_GBRANCH

### DIFF
--- a/src/regexpr.pas
+++ b/src/regexpr.pas
@@ -1715,7 +1715,10 @@ const
   OP_SUBCALL = TREOp(65); // Call of subroutine; OP_SUBCALL+i is for group i
   OP_LOOP_POSS = TREOp(66); // Same as OP_LOOP but in non-greedy mode
 
-  OP_GBRANCH = TREOp(67); // Guarded branch
+  // Guarded branch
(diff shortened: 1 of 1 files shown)


Alexey-T:
Can you write in the src, what is the OP_GBRANCH? how it differs from OP_BRANCH?

Alexey-T:
Does it speedup something?

User4martin:
> Does it speedup something?

Yes. gets my current work from 1,25 to 1.20 secs. That's 4%

Alexey-T:
can you decipher the strange comment? -
```
begin // Emit x* as (x&|), where & means "self".
```

User4martin:
Before I make a commit

```
            // To complex for OP_STAR. Write loop using OP_BRANCH
            // 1: loop-content
            // 2: OP_Branch with 2 branches - to allow backtracking
            // 2a: OP_BACK(1) to match the loop again
            // 2b  OP_NOTHING to exit, if the loop can match no more
```

verbose enough?

Alexey-T:
yes. good.

Alexey-T:
what does it mean: `2a: OP_BACK(1)` ?
what is `(1)`.

Issue #10, opened 2019-06-06
 UNHANDLED EXCEPTION!!! TRegExpr(comp): Urecognized Modifier (pos 10)


chevant0n:
^(?!([0-9])\1{9})[0-9]{10}$

1111111111 - not ok
1111111112 - ok

andgineer:
It does not support look-ahead
https://regex.sorokin.engineer/en/latest/faq.html#regular-expressions-with-do-not-work

andgineer:
Duplicates
https://github.com/andgineer/TRegExpr/issues/1

PR #378, opened 2024-01-08
Small optimization / Needs testing with Delphi

User4martin:
The first commit is ready to go. Just skipping unnecessary mem alloc.

----

**The 2nd commit 6438a0 needs some testing with Delphi.**

It replaces `GrpBounds[regRecursion].GrpStart` with direct pointe access. Skipping the need to calculate the element for `regRecursion` (which in 99% of cases is zero anyway).

On my PC, the benchmark gains 2%  // Real live apps gain less, but still.

I know Delphi should have "PointerMath". I don't know if it is the same compiler directive. And I don't know if it works 100% identical. So that needs testing.

I tried to just use an `array of PRegExprChar` instead of `^PRegExprChar`. But that slowed things down. It adds refcounts, and then, if one array is resized, the other isn't...

The pointers are safe. They are only used in MatchPrim. Nothing can change the underlying arrays during the run. And `ClearInternalExecData` ensures the value are valid at start.
Data is still stored in the array, and any access outside MatchPrim reads it from the arrays, and not via the pointers.

### DIFF
--- a/src/regexpr.pas
+++ b/src/regexpr.pas
@@ -397,6 +397,7 @@ type
     // work variables for compiler's routines
     regParse: PRegExprChar; // pointer to currently handling char of regex
     regNumBrackets: Integer; // count of () brackets
+    regNumAtomicBrackets: Integer; // count of (?>) brackets
     regDummy: array [0..8 div SizeOf(REChar)] of REChar; // dummy pointer, used to detect 1st/2nd pass of Compile
                       // if p=@regDummy, it is pass-1: opcode memory is not yet allocated
     programm: PRegExprChar; // pointer to opcode, =nil in pass-1
@@ -565,7 +566,7 @@ type
 
     // regular expression, i.e. main body or parenthesized thing
     function ParseReg(InBrackets: Boolean; var FlagParse: Integer): PRegExprChar;
-    function DoParseReg(InBrackets, IndexBrackets: Boolean; var FlagParse: Integer; BeginGroupOp, EndGroupOP: TReOp): PRegExprChar;
+    function DoParseReg(InBrackets: Boolean; BracketCounter: PInteger; var FlagParse: Integer; BeginGroupOp, EndGroupOP: TReOp): PRegExprChar;
 
     // one alternative of an | operator
     function ParseBranch(var FlagParse: Integer): PRegExprChar;
@@ -3194,6 +3195,7 @@ begin
     fCompModifiers := fModifiers;
     regParse := ARegExp;
     regNumBrackets := 1;
+    regNumAtomicBrackets := 0;
     regCodeSize := 0;
     regCode := @regDummy;
     regCodeWork := nil;
@@ -3201,6 +3203,7 @@ begin
     EmitC(OP_MAGIC);
     if ParseReg(False, FlagTemp) = nil then begin
       regNumBrackets := 0; // Not calling InitInternalGroupData => array sizes not adjusted for FillChar
+      regNumAtomicBrackets := 0;
       Exit;
     end;
 
@@ -3213,6 +3216,7 @@ begin
     fCompModifiers := fModifiers;
     regParse := ARegExp;
     regNumBrackets := 1;
+    regNumAtomicBrackets := 0;
     GrpCount := ParsedGrpCount;
     ParsedGrpCount := 0;
     regCode := programm;
@@ -3314,10 +3318,10 @@ end; { of function TRegExpr.CompileRegExpr
 
 function TRegExpr.ParseReg(InBrackets: Boolean; var FlagParse: Integer): PRegExprChar;
 begin
-  Result := DoParseReg(InBrackets, True, FlagParse, OP_OPEN, OP_CLOSE);
+  Result := DoParseReg(InBrackets, @regNumBrackets, FlagParse, OP_OPEN, OP_CLOSE);
 end;
 
-function TRegExpr.DoParseReg(InBrackets, IndexBrackets: Boolean;
+function TRegExpr.DoParseReg(InBrackets: Boolean; BracketCounter: PInteger;
   var FlagParse: Integer; BeginGroupOp, EndGroupOP: TReOp): PRegExprChar;
 // regular expression, i.e. main body or parenthesized thing
 // Caller must absorb opening parenthesis.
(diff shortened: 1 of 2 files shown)


User4martin:
Added some more commits.

**Depending on the outcome of Delphi tests (and/or review, if there is a matter) I can re-arrange the list of commits according to requests, before they get merged.**

User4martin:
- `(?i:..)` quantifiable => This is a bugfix. In the version with the colon, the group actually matches something. So it can be quantified (just like `(?:...)`

- No OP_OPEN for `(?i:...)` improves speed... Well if there is heavy use of such groups. Usually they don't happen often. But it was also needed for the GrpIndexes removal.

- Separate counters: Less mem alloc, since each type of group needs different data, and previously room for both types of data was allocated for both types of groups

- Remove GrpIndexes: General clean up. Maybe (not tested) tiny speed improvement for compilation (not execution)

Alexey-T:
I dont have last Delphi. also I am on Linux.
I tag users which were having Delphi, please test Martin's diff:
 @dkounal  @AmigoJack  @dummzeuch  @HPWi

dkounal:
Tested with Delpi 11.3 
test\test_delphi: needs the following in tests.pas to run:
const LineEnding = #10;
type SizeInt = LongInt;

but it stops with the following error in line 8432:
'TRegExpr compile: unknown meta-character: \p (pos 13)' 

Screen results:
**********************************************************************
*        DUnitX - (c) 2015-2018 Vincent Parrett & Contributors       *
*                                                                    *
*        License - http://www.apache.org/licenses/LICENSE-2.0        *
**********************************************************************

DUnitX - [test_delphi.exe] - Starting Tests.

...................................E.................................................................F.................................................E.E..............................................

Tests Found   : 100
Tests Ignored : 0
Tests Passed  : 96
Tests Leaked  : 0
Tests Failed  : 1
Tests Errored : 3

Failing Tests

  tests.TTestRegexpr.RunTest27
  Message: Search position, expected: <1> but was: <-1>
Tests With Errors
  tests.TTestRegexpr.TestIsFixedLength
  Message: TRegExpr compile: unknown meta-character: \p (pos 13)

  tests.TTestRegexpr.RunTest51unicode
  Message: TRegExpr compile: unknown meta-character: \p (pos 1)

  tests.TTestRegexpr.RunTest52unicode
  Message: TRegExpr compile: unknown meta-character: \p (pos 2)

dkounal:
I was some commits behind, so I updated and running tests before the above commit is the following:
**********************************************************************
*        DUnitX - (c) 2015-2018 Vincent Parrett & Contributors       *
*                                                                    *
*        License - http://www.apache.org/licenses/LICENSE-2.0        *
**********************************************************************

DUnitX - [test_delphi.exe] - Starting Tests.

.....................................E.................................................................F.................................................E.E.....................................F........

Tests Found   : 101
Tests Ignored : 0
Tests Passed  : 96
Tests Leaked  : 0
Tests Failed  : 2
Tests Errored : 3

Failing Tests

  tests.TTestRegexpr.RunTest27
  Message: Search position, expected: <1> but was: <-1>

  tests.TTestRegexpr.RunTest70
  Message: Replace failed, expected: <abc> but was: <abr#0c>



Tests With Errors

  tests.TTestRegexpr.TestIsFixedLength
  Message: TRegExpr compile: unknown meta-character: \p (pos 13)

  tests.TTestRegexpr.RunTest51unicode
  Message: TRegExpr compile: unknown meta-character: \p (pos 1)

  tests.TTestRegexpr.RunTest52unicode
  Message: TRegExpr compile: unknown meta-character: \p (pos 2)

After the above commit:
**********************************************************************
*        DUnitX - (c) 2015-2018 Vincent Parrett & Contributors       *
*                                                                    *
*        License - http://www.apache.org/licenses/LICENSE-2.0        *
**********************************************************************

DUnitX - [test_delphi.exe] - Starting Tests.

.....................................E.................................................................F.................................................E.E..............................................

Tests Found   : 101
Tests Ignored : 0
Tests Passed  : 97
Tests Leaked  : 0
Tests Failed  : 1
Tests Errored : 3

Failing Tests

  tests.TTestRegexpr.RunTest27
  Message: Search position, expected: <1> but was: <-1>



Tests With Errors

  tests.TTestRegexpr.TestIsFixedLength
  Message: TRegExpr compile: unknown meta-character: \p (pos 13)

  tests.TTestRegexpr.RunTest51unicode
  Message: TRegExpr compile: unknown meta-character: \p (pos 1)

  tests.TTestRegexpr.RunTest52unicode
  Message: TRegExpr compile: unknown meta-character: \p (pos 2)

Alexey-T:
`'TRegExpr compile: unknown meta-character: \p (pos 13)'`
occurs  if RegEx engine did not have the define `UnicodeRE`. it is On by default in regexpr.pas.

```
// ======== Define options for TRegExpr engine
{$DEFINE UnicodeRE} // Use WideChar for characters and UnicodeString/WideString for strings   
```

User4martin:
Actually I just looked, `\p` depends on
```
{$IFDEF FPC}
  {$DEFINE FastUnicodeData} // Use arrays for UpperCase/LowerCase/IsWordChar, they take 320K more memory
{$ENDIF}
```

Not sure why this is FPC only?

But it was made so in
Revision: 85b20bde027d0759ad93a6a745b0fc5c8c7a1b58
Author: Michael Van Canneyt 
Date: 19/02/2021 11:31:45
* Use FastUniCodeData by default only in FPC

Though before that, it was not enabled by default. 

It was made non-default in
Revision: 861691f354ea032e90e0026a312111db4f964be4
Author: Michael Van Canneyt
Date: 19/02/2021 11:21:36
* Fix compilation for D2009> when no FastUniCodeData is defined

Alexey-T:
>Not sure why this is FPC only?

Because Delphi has TCharacter which does the job and it is better than our array-method.

User4martin:
Some more changes:

- Values like "AOffset" only need to be checked if they are user specified.
- Other checks can be done before the try/except block
- No idea why I had the Error re-raised when the option was set NOT to raise it.
- IsPartFixedLength did sometimes give up early / values were correct but in the sense that the MinLen was reported smaller than it actually was => so all code would work (since more than minlen was matched, and that is a possible outcome), but sometimes do more work than needed.

User4martin:
> So, i can apply the change? no errors related to it?

Yes.

The only question was if the PointerMath works. 

It does, otherwise it would likely not have compiled, but if it had compiled other tests would have failed (recursion tests)

User4martin:
- When text was set with `SetInputSubString` it may not end with #0. So FirstCharSet does not work.

PR #392, opened 2025-02-05
Fix #391

### LINKED ISSUE #391: SubExprMatchCount allways max. SubExprCount
SubExprMatchCount is allways max. SubExprCount but not lower if not all Groups are matched.

Examples (from the doku of the property):
  // For example: Expression := '(1)?2(3)?';
  //  Exec ('12'): SubExprMatchCount=1, Match[0]='12', [1]='1'
RE.SubExprMatchCount = 2 <--- should be 1

  //  Exec ('2'): SubExprMatchCount=0, Match[0]='2'
RE.SubExprMatchCount = 2 <--- should be 0

horst-p-w-neubauer:


horst-p-w-neubauer:
It's in essence your old code (like in 0.952 - my refference) for the new internal structure including your fix for result:=-1 when the exec() fails.

horst-p-w-neubauer:
sorry, I thougt I was working on the trunk, but was on 1.182

horst-p-w-neubauer:
I see, 1.189 doesn't compile with my Delphi 2007 Compiler. You try to set compiler directive $PointerMath from D8 up but it was introduced first in Delphi 2009 - when I'm right (https://blog.therealoracleatdelphi.com/2008/01/taking-big-pbyte-of-pointer-math_24.html)

Alexey-T:
I fixed condition for $PointerMath, pls check again.

Alexey-T:
Adapt your diff for my new fixed repo with $PointerMath.

```
  if Length(GrpBounds[0].GrpStart) = 0 then begin
    Exit;
  end;
  // if nothing found, we must return -1 per TRegExpr docs
  if (GrpBounds[0].GrpStart[0] <> nil) then begin
    Result := Length(GrpIndexes)-1;
    Result := Length(GrpBounds[0].GrpStart)-1;
    while (GrpBounds[0].GrpStart[ Result ] = nil) do begin
      Dec( Result );
    end;
```

- remove redundant begin/end braces
- ` Result := Length(GrpBounds[0].GrpStart)-1;` feels wrong - you assign Length-1 to Count, while usually we assign Length to Count.

Alexey-T:
thanks for detailed answer.
feels good. do you agree @user4martin ?

Alexey-T:
yes, existing apps may be broken now, so let's rename new working code to property `SubExprMatchedCount` (example name).

can you also add your 2 tests to tests-project (folder 'tests')?

horst-p-w-neubauer:
I'll try to.

horst-p-w-neubauer:
I'm not sure, if renaming is such a good idea. 
Infact: the behavior changed from 0.952 to 1.18x as well, thats why **my** code was broken and I got engaged in your project. 

It's just changing back to the defined and former behavior in case of matching inputs and repairs the old wrong behavior for not matching inputs (was 0 instead of -1).

Is the new test what you've had in mind?

Alexey-T:
I got your repo (exact branch with change) and run 'tests' proj. 
fails on N tests now!

Снимок экрана 2025-02-05 183749

horst-p-w-neubauer:
That's what I've tried to say above.

Alexey-T:
yes, I did not checks tests before. I must apply only 'passed tests'.

horst-p-w-neubauer:
I try to. Probably it's solved if you remove the -1,-1 pairs in IsMatching - I assume, you've added them to solve the wrong behavior of SubExprMatchCount.

horst-p-w-neubauer:
It solved some of the problems - but for the others there has to look someone who better understands the RegEx that is tested.

Alexey-T:
so this patch makes some old code broken.
because SubExprMatchCount result is decreased for some expressions.

is it decreased here correctly? ->

Снимок экрана 2025-02-05 183749

I cannot fully understand this test and all other @user4martin 's tests, so not sure.

also @USER4martin help is needed to correct other tests.

horst-p-w-neubauer:
I'm sure with the atomic tests. 
The corrected behavior is, that SubExprMatchCount only counts to the last found group not to the max. possible groups. 
The trailing -1,-1 pairs are filling up until the max. possible groups. That is no longer needed.

User4martin:
> is it decreased here correctly?

> The corrected behavior is, that SubExprMatchCount only counts to the last found group not to the max. possible groups.
> The trailing -1,-1 pairs are filling up until the max. possible groups. That is no longer needed.


Ok, IMHO more a question of, is this what it should do. (as in "should do in future" rather than "has at some point in the past been doing"

Its for Alexey to define which behaviour he wants from that property. Both can be used by code using TRegExpr.

See my question and example on the initial report https://github.com/andgineer/TRegExpr/issues/391#issuecomment-2637366241 (consider it answered by your statement).

Reducing the count for some non-matches, but not for other non-matches seems arbitrary to me. But again Alexey to decide what he wants from it.

If SubExprMatchCount is not giving the total count, is there another property that will? For anyone who needs that (and is currently relaying on SubExprMatchCount  to get it / I have not checked if there is)


> Infact: the behavior changed from 0.952 to 1.18x as well, thats why my code was broken and I got engaged in your project.

Do we know which commit changed that. Maybe the commit states 
- something that was fixed by it, and needs to be tested now, maybe fixed different
- that the change was an intentional change of the behaviour
- that the change was an unwanted side effect

User4martin:
It is indeed documented like that in the comment
```
    // Number of subexpressions has been found in last Exec* call.
    // If there are no subexpr. but whole expr was found (Exec* returned True),
    // then SubExprMatchCount=0, if no subexpressions nor whole
    // r.e. found (Exec* returned false) then SubExprMatchCount=-1.
    // Note, that some subexpr. may be not found and for such
    // subexpr. MathPos=MatchLen=-1 and Match=''.
    // For example: Expression := '(1)?2(3)?';
    // Exec ('123'): SubExprMatchCount=2, Match[0]='123', [1]='1', [2]='3'
    // Exec ('12'): SubExprMatchCount=1, Match[0]='12', [1]='1'
    // Exec ('23'): SubExprMatchCount=2, Match[0]='23', [1]='', [2]='3'
    // Exec ('2'): SubExprMatchCount=0, Match[0]='2'
    // Exec ('7') - return False: SubExprMatchCount=-1
```

So the change may have been non-intentional

User4martin:
Found it. This is the commit that removed the behaviour initially (At least I think).

SHA-1: 13f826c6496f34af57304f5872725bc3251b7e8f
* optimize speed of non-capturing groups storing

Alexey-T:
Martin, i agree, that was intentional change. Lets keep this change.

If users need property to count only existing/matching groups, we can add different property.

User4martin:
I also think the change is for the better.

But besides this, in the patch I found
Old: `   Result := GrpCount;`
New: `    Result := Length(GrpBounds[0].GrpStart)-1;`

But the length may be bigger than GrpCount. See
```
  // Increment/keep-capacity for the size of arrays holding 'Group' related data
  // e.g., GrpBounds, GrpOpCodes and GrpNames
  RegexGroupCountIncrement = 50;
```

Which means for speed sake the array is not resized (shrunk), if the savings would only be small. Especially because the next pattern to be matched may have to grow it again, if the Expression is changed to something with more groups)

So, if that had needed to be changed back, then it would need to count down from GrpCount.

horst-p-w-neubauer:
> So, if that had needed to be changed back, then it would need to count down from GrpCount.
good point, I'll take this in

horst-p-w-neubauer:
this even solved the most of the broken tests with only two remaining (except the two last that where broken bevore the fix as well)
grafik

Is anybody able to solve TestLoop and TestEmptyLoop? I'm not and I actually have to leave for today as I'm expected to attend a well loved parent's evening at school.

User4martin:
As indicated, in generic I see no advantage of reducing `SubExprMatchCount`. As there can be gaps in the middle, any code using it would still need to check.

Well, except some very special cases. If you have a regex where only the last group is optional `(...)?`. Or if you know for certain that if one optional grp was not matched, then all other grp after it also were not matched. Then, you could indeed use the reduced number as quick indicator, and save yourself from checking the entries.

I don't think we should change `SubExprMatchCount` back to cover this special case (as this may break other code relying on the new behaviour).

But, if Alexey agrees, I could imagine a new property (better names welcome...). Something like either
- `IndexOfHighestMatchingGroup` / `HighestMatchingGroup` / `HighestMatchingGroupIndex` => what you seek
- or `MatchingGroupCount`  => also taking out non-matches in the middle

horst-p-w-neubauer:
But there is nothing like a a special case here! It's just restoring the ever documented behavior of this property. Why should anyone change this just for it has been broken for a little while?

User4martin:
Just to clarify: I do not call the decision on if to keep or revert. I am only on board for the review on side effects (like the GrpCount issue).

On the desired behaviour, my comments are merely my opinion (I,e, if I were the main maintainer, but I am not). So on that it is for Alexey to make the call.

----

If he is happy with changing the behaviour as you suggest, then I would suggest to amend `TTestRegexpr.IsMatching` to detuct trailing -1 from the count. 

That way the -1 pairs could remain in the test (at least most of them probably). The advantage would be, that if ever a "AllGroupCount" was introduced, the numbers for testing this would be in the testcase.

But, IMHO it is not a requirement for an acceptable pull request. So since you already removed many of them, if the remaining can be fixed easier in a way that does not modify TTestRegexpr.IsMatching then that way would be ok. If TTestRegexpr.IsMatching  needs the modification, then IMHO keep the -1 where possible.

Alexey-T:
I like the idea if patch, so will apply.

I will see why some tests do not work, maybe define for Unicode regex is not enabled in the regexpr.pas?

Alexey-T:
@horst-p-w-neubauer I suggest to revert removing of -1,-1 ending pairs in tests, and to pass the tests - i suggest to edit IsMatching. it must compare calculated length not with SubExprMatchCount but with SubExprAllCount  --> make new property SubExprAllCount.
please.
this way we will keep existing tests.

horst-p-w-neubauer:
In the case of the last two failing test, please keep in mind, that I'm
developing in Delphi 2007 - this tests might not been designed for such an
old environment.

Alexey T. ***@***.***> schrieb am Mi., 5. Feb. 2025, 20:57:

Alexey-T:
Ok, but i will test it on FPC.

horst-p-w-neubauer:
Do I understand you right, that we plan to keep the behavior of `SubExprMatchCount` as commented in the proeprty description and we add a new propeperty `SubExprAllCount? This new property should than be used to restore the test implementations that I changed today. 

I would absolute be fine with that.

Who would write the description for the new property and therefore define its behavior? 
Is this new property any longer dependend on the `Exec()` call or is it a result of the `Expression` parsing? If I understand the code right than `GrpCount=0 or GrpCount=ParsedGrpCount` and never gets `ParsedGrpCount<0`. So I guess `SubExprAllCount` will never be `-1` as `SubExprMatchCount` will become in case of `Exec()=false`.

Alexey-T:
- SubExprMatchCount - new work of property, it trims last empty items. it keeps -1 result for none matches as before. It must be NOT used in 'tests'
- SubExprAllCount - new name for old property. it must be used in 'tests' for last -1,-1 pairs.

horst-p-w-neubauer:
Beside our discussion I've found, that the If in front of the while loop is not neccessary if we test for Result <= 0

horst-p-w-neubauer:
I wonder if it would be helpfull to distinguish between Groups as a result of the Expression parsing and Matches as a result of the Exec() call. Wouldn't that lead us to `property SubExprGroupCount`  instead of `SubExprAllCount `?

horst-p-w-neubauer:
I see. So SubExprAllCount is nothing else than the existing GroupCount. So maybe that should be the name.

Alexey-T:
With last patch, tests fail:
`Empty group: ?? Match count`

Code is here
```
      ExpGrpPos1 := 1;
      ExpGrpLen1 := 0;
      if MatchZeroTimes then begin
        ExpGrpPos1 := -1;
        ExpGrpLen1 := -1;
      end;

      s := LoopTxt + ' / ' + PtnText;
      IsMatching('Empty group '+s,                PtnText+LoopTxt,      'a',   [1,0,  ExpGrpPos1,ExpGrpLen1]);
```

User4martin:
As for the test "empty group"

```
diff --git a/test/tests.pas b/test/tests.pas
index 8afd5c7..a3e901e 100644
--- a/test/tests.pas
+++ b/test/tests.pas
@@ -892,7 +892,9 @@ begin
 
   IsTrue(AErrorMessage + ' Exec must give True', RE.ExecPos(AOffset, AMustMatchBefore));
 
-  L := Length(AExpectStartLenPairs) div 2;
+  L := Length(AExpectStartLenPairs);
+  while (L > 1) and (AExpectStartLenPairs[L-2] = -1) do dec(L,2);
+  L := L div 2;
   AreEqual(AErrorMessage + ': MatchCount', L - 1, RE.SubExprMatchCount);
   for i := 0 to L - 1 do begin
     AreEqual(AErrorMessage + ': MatchPos['+inttostr(i)+']', AExpectStartLenPairs[i*2], RE.MatchPos[i]);
```

And then you can (probably / I have not tested) restore all the removed "-1,-1". 

So they will be available for testing a future "[All/Parsed/Full]GroupCount"

In fact, the loop at the end, could run to the original L. So maybe introduce `MatchedCnt := L;` and then reduce that in the while loop. (and both MatchedCnt and L will then need div 2)

Alexey-T:
so hard epopea/sage. to apply tiny change. I now don't like this idea. this change trims trailing non-matched groups but don't remove middle groups.

horst-p-w-neubauer:
I'm a bit confused. I've had a very busy week in my job so I could not attent to this discussion for some days. Would closing this pullrequest mean, You stay with your bug?

I even can't understand your problem. This talk of "empty middle groups" is nonsense. The purpose of this property has ever been the same: give me the last matched group! Regardless of any empty middle groups or the complete possible number of groups. That's how You've commented this property when it was envoked until today. But You've never implemented it bugfree. When I now provide You with a working implementation You throw it away, because - well - I actually can't see any serious reason. Some omnious "it may break someones code..". Yes. The shift in the bug-implemention broke my code. As far as I know is this the only known code that is broken. Sorry, I've had a lot of pressure this week and this bug was one of the reasons.

Alexey-T:
Maybe you are right. I will see the patch again.

horst-p-w-neubauer:
I'm absolutely willing to help fix all the problems that comes with my work. But I'm blocked for the next one or two weeks because we **have** to develiver a testsystem and provide narrow supporting.

Alexey-T:
@User4martin sorry for 3 week delay.
I got the Git branch from @horst-p-w-neubauer (fix-issue-391 branch) and tests project now fails.
here: TestEmptyLoop function.
that was your addition AFAIR. so can you please get this branch, and correct the test if needed?

User4martin:
See my reply here:
https://github.com/andgineer/TRegExpr/pull/392#issuecomment-2642692651

That patch is still fixing the test case.

And, additionally, when that patch is applied it should (I have not tested) be possible to keep all the `-1,-1` pairs in the expectations.
E.g. the following removal (that is currently part of the pull request) should not be needed:
```
@@ -1227,7 +1234,7 @@ begin
              '1ab__cx__9cx__8...56Yb');
   IsMatching('NESTED: Atomic backtrace inner, outer can still try and find ',
              'a(?>((?>b.*?cx..8|.)_)|.)',
-             '1ab__cx__9cx__8...56Yb',  [2,2,  -1,-1]);
+             '1ab__cx__9cx__8...56Yb',  [2,2]);
 
```

Keeping the `-1,-1` means that they can later be used, if a property is added that returns the parsed count of groups in the expression (rather than the number of the highest matching group).

Alexey-T:
Applied by hands, without all 'tests' changes - we don't need to change 'tests' with @User4martin patch.

